PR #43, opened 2019-08-12
Feature/docs

rfaircloth-splunk:


docs/index.md
```diff
@@ -0,0 +1,23 @@
+# Welcome to Splunk Connect for Syslog
+
+Splunk Connect for Syslog is an open source packaged solution for 
+getting data into Splunk using syslog-ng (OSE) and the Splunk
```
mkarlstrand-splunk: Spell out “Open Source Edition” for first mention along with “OSE” abbreviation, then use abbreviation after that

rfaircloth-splunk: done

docs/index.md
```diff
@@ -0,0 +1,23 @@
+# Welcome to Splunk Connect for Syslog
+
+Splunk Connect for Syslog is an open source packaged solution for 
+getting data into Splunk using syslog-ng (OSE) and the Splunk 
+HTTP event Collector. 
+
+## Project Goals
+
+* Bring a tested configuration and build of syslog-ng OSE to the market that will function consistently regardless of linux distribution
```
mkarlstrand-splunk: Can the solution really support any type of Linux? Even if true I think it might better to state this more conservatively. "validated to function consistently on RedHat Base Image, etc, etc." 
So the scope is defined.

rfaircloth-splunk: our regardless is limited to the set of distros supported by docker so we are fairly safe. That is actually what we mean their host can be Suse or RHEL we can give the syslog as an app container no matter what their standard base is

docs/gettingstarted.md
```diff
@@ -0,0 +1,58 @@
+
+# Pre-req
+
+* Linux host with Docker 19.x or newer with Docker Swarm enabled
+    [Getting Started](https://docs.docker.com/get-started/)
+* A Splunk index for metrics typically "em_metrics"
```
mkarlstrand-splunk: Please clarify a bit. Does this mean the Admin has to create an "em_metrics" index or can they expect that this is already there for other reasons?

docs/gettingstarted.md
```diff
@@ -0,0 +1,58 @@
...
+* One or more Splunk indexes for events collected by SC4S
+* Splunk HTTP event collector enabled with a token dedicated for SC4S
+    * [Splunk Enterprise](http://dev.splunk.com/view/event-collector/SP-CAAAE6Q)
+    * [Splunk Enterprise Cloud](http://docs.splunk.com/Documentation/Splunk/7.3.1/Data/UsetheHTTPEventCollector#Configure_HTTP_Event_Collector_on_managed_Splunk_Cloud)
+* A network load balancer (NLB) configured for round robin. Note: Special consideration may be required when more advanced products are used. The optimal configuration of the load balancer will round robin each post (not each connection)
```
mkarlstrand-splunk: "round robin each post"
Should this be "port" ?

docs/gettingstarted.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Setup
+
+* Create a directory on the server for configuration
```
mkarlstrand-splunk: Should we suggest a naming for this directory?

docs/gettingstarted.md
```diff
@@ -0,0 +1,58 @@
...
+* Create a docker-compose.yml file based on the following template
```
mkarlstrand-splunk: Would it be easier if we provided a file they can use as a template vs c/p this text?

docs/gettingstarted.md
mbonsack: I believe "post" is correct; the round robin applies to each http post request made by the client.

rfaircloth-splunk: no post,  a connection and accept multiple posts while open but with ALB the RR is at the post level

docs/gettingstarted.md
rfaircloth-splunk: "em_metrics" is quasi standard the IAI and ITSI apps use it

rfaircloth-splunk:
> disk-buf-size(20000) is still too small; think you want (200000). Was this a regression/typo?

omission but thats a bug for another pr

docs/gettingstarted.md
rfaircloth-splunk: downloading from github individual files harder than it should be CP is easier and never breaks company rules about downloading from the internet

docs/gettingstarted.md
mbonsack: Admin must create it.  Unlike "main" (which is created automatically for event data), there is no automatically-created, default index for metrics.

docs/index.md
```diff
@@ -0,0 +1,24 @@
+# Welcome to Splunk Connect for Syslog
+
+Splunk Connect for Syslog is an open source packaged solution for 
+getting data into Splunk using syslog-ng open source edition (Syslog-NG OSE) and the Splunk 
+HTTP event Collector. 
+
+## Project Goals
+
+* Bring a tested configuration and build of syslog-ng OSE to the market that will function consistently regardless of linux distribution
```
mbonsack:
```suggestion
* Bring a tested configuration and build of syslog-ng OSE to the market that will function consistently regardless of the underlying host's linux distribution
```

docs/index.md
```diff
@@ -0,0 +1,24 @@
+# Welcome to Splunk Connect for Syslog
+
+Splunk Connect for Syslog is an open source packaged solution for 
+getting data into Splunk using syslog-ng open source edition (Syslog-NG OSE) and the Splunk
```
mbonsack:
```suggestion
getting data into Splunk using syslog-ng Open Source Edition (Syslog-NG OSE) and the Splunk 
```

docs/gettingstarted.md
```diff
@@ -0,0 +1,58 @@
+
+# Pre-req
+
+* Linux host with Docker 19.x or newer with Docker Swarm enabled
+    [Getting Started](https://docs.docker.com/get-started/)
```
mbonsack:
```suggestion
    * [Getting Started](https://docs.docker.com/get-started/)
```

mbonsack: Forgot bullet; renders improperly

docs/gettingstarted.md
mbonsack:
```suggestion
* A network load balancer (NLB) configured for round robin. Note: Special consideration may be required when more advanced products are used. The optimal configuration of the load balancer will round robin each http POST request (not each connection)
```

mbonsack: A network load balancer should _not_ be required; internal syslog-ng LB works well and should be encouraged if the customer situation will allow it.  This will allow proper, keep-alive sessions directly to each indexer to be configured, at the expense of some overhead in properly maintaining the indexer list.

docs/performance.md
```diff
@@ -0,0 +1,28 @@
...
+```
+average rate = 9717.58 msg/sec, count=1749420, time=180.026, (average) msg size=800, bandwidth=7591.86 kB/sec
+
+```
+
+![Splunk Metrics](balanced.png "Events balanced across 5 indexers")
+
+## Limitations
+
+* Splunk Enterprise's implementation of the http event collection server will respond to the client with a status code 200 and fail to commit the events to disk during a rolling restart in our testing 20-30 events per indexer pay be lost
```
mbonsack:
```suggestion
* Splunk Enterprise's implementation of the http event collection server will respond to the client with a status code 200 and fail to commit the events to disk during a rolling restart in our testing 20-30 events per indexer may be lost
```

docs/gettingstarted.md
rfaircloth-splunk: I agree that's actually true if we allow non LB use then we should require DNS a-records where "hec-splunk.company.com" has all the entries for all the indexers. I don't think we should maintain the list in docker (I already tested this it works)

docs/gettingstarted.md
rfaircloth-splunk: no I don't think we should

package/etc/conf.d/destinations/d_destinations.conf
```diff
@@ -17,15 +17,16 @@ destination d_hec {
...
          batch-timeout(3)
          timeout(15)
          user_agent("syslog-ng User Agent")
          user("syslog-ng")
+         headers("Connection: close")
          password("`SPLUNK_HEC_TOKEN`")
          persist-name("splunk")
-         disk-buffer(mem-buf-length(500)
+         disk-buffer(mem-buf-length(15000)
                      disk-buf-size(20000)
```
mbonsack: This buffer size is too small; thought I previously commented on this...

PR #82, opened 2019-09-13
Fix/dynamic template

mbonsack:
Add Dynamic template capability
Add new t_JSON_5424 template for RFC5424-compliant messages
Add default source metadata

### DIFF
--- a/package/etc/conf.d/conflib/_common/templates.conf
+++ b/package/etc/conf.d/conflib/_common/templates.conf
@@ -44,13 +44,28 @@ template t_hdr_sdata_msg {
         };
 
 # ===============================================================================================
-# JSON; for JSON pretty-printing
+# JSON; for JSON pretty-printing (for debugging)
 # ===============================================================================================
- 
+
 template t_JSON {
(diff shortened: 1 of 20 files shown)


package/etc/conf.d/conflib/_common/templates.conf
```diff
@@ -44,13 +44,28 @@ template t_hdr_sdata_msg {
...
+	};
+
+# ===============================================================================================
+# JSON; for JSON pretty-printing (for RFC5424)
+# ===============================================================================================
+
+template t_JSON_5424 {
+	template("$(format-json --scope all-nv-pairs
+	            --exclude fields.*
+	            --exclude .splunk.*
```
rfaircloth-splunk: should we exclude all .* fields

mbonsack: I think we want all fields except those that will be represented by indexed fields in Splunk.  The only change in this template vs. the regular one is the exclusion of $RAWMSG.  You don't want to exclude all .* fields because SDATA is represented that way...

PR #97, opened 2019-09-27
Feature/proofpoint

mbonsack:
Add filter and log paths for Proofpoint PPS filter events and sendmail events per existing TA


package/etc/conf.d/filters/proofpoint/proofpoint.conf
```diff
@@ -0,0 +1,18 @@
+# Proofpoint
```
rfaircloth-splunk: Rename this file to pps.conf

mbonsack: Done; changed to pps.conf

PR #109, opened 2019-10-02
removing optional parts

l-kutch:
Since some steps were optional, I didn't do them :) So I ran into errors. Revising to make the parts mandatory.


l-kutch:
Can it be merged as-is for now? I don't want it to get old & conflicty, as I might not be making the next round of changes till 10/9. Thanks.

PR #170, opened 2019-10-18
Provide documentations for K8S at the edge using K3s

rfaircloth-splunk:


rfaircloth-splunk:
This PR is presently blocked due to inability to accept connection on 514 for both TCP and UDP in single node K8s clusters

PR #184, opened 2019-11-05
Add disk buffer envrionment variables table

mbonsack:
Add env var table to docs
Adjust default disk buffer to 50 GB


package/etc/conf.d/destinations/splunk_hec.conf.tmpl
```diff
@@ -25,7 +25,7 @@ destination d_hec {
             reliable(no)
             {{- end}}
 
-             disk-buf-size({{- getenv "SC4S_DEST_SPLUNK_HEC_DISKBUFF_DISKBUFSIZE" "1048576"}})
+             disk-buf-size({{- getenv "SC4S_DEST_SPLUNK_HEC_DISKBUFF_DISKBUFSIZE" "53687091200"}})
```
themrkeys: this has to be math each worker thread has to divide and round nearest int

mbonsack: Fix added for above.

PR #187, opened 2019-11-07
Support K8S container logging

rfaircloth-splunk:
This PR will support syslog-ng as the log collection tool in place of fluentd


perditi0nspam:
Any news when there may be support for Openshift/K8's? The FAQ was recently updated to explain a few reasons why it shouldn't be installed on these platforms, but not any kind of answer regarding if these shortcomings will be addressed by Redhat.

PR #259, opened 2020-01-03
Update sc4s_vendor_product values

mbonsack:
* Provide more granular `sc4s_vendor_product` values to help distinguish the actual log path used.
* Results in far cleaner archive files on disk


mbonsack:
Changed the `test_tag` test to `test_fallback` and removed check for specific `sc4s_vendor_product` value.

mbonsack:
Re: sc4s_vendor_product docs:  The values for this field are not documented at all. Keep in mind the key for `splunk_indexes.conf` and `sc4s_vendor_product` values are distinct.  I have tried to align them with the keys in `splunk_indexes.conf`, but there is no requirement that they be the same at all (and in some cases they do differ).  The value for `sc4s_vendor_product` just needs to match any corresponding filter(s), and is also provided as an indexed field in Splunk to indicate what log path the message traversed.

PR #262, opened 2020-01-06
Gomplate update/cleanup for "soup" port defaults

mbonsack:
* Add defaults in template for `source_network.t`
* Remove default listening ports in entrypoint.sh and docker-compose.yml files
* Clean up all log path templates to make final output more readable
* Minor cleanup to `network.conf.tmpl` for readability


mbonsack:
* Added check for TLS enable env variable
* Remove extraneous `conv.ToBool` declarations in 3 log path files

PR #297, opened 2020-01-31
Document non root mode

rfaircloth-splunk:


rfaircloth-splunk:
This was added back in #298 so this is redundant

PR #323, opened 2020-02-12
FortiWeb initial commit

bkresoja:
FortiWeb detection and classification


rfaircloth-splunk:
@bkresoja I made some minor updates, we made some general improvements to the test cases and I applied those to you changes great work. See #325 for the result

PR #340, opened 2020-03-11
For Palo Alto networks devices use the device_name field as hostname

rfaircloth-splunk:


mbonsack:
Check
https://docs.paloaltonetworks.com/pan-os/7-1/pan-os-admin/monitoring/use-syslog-for-monitoring/syslog-field-descriptions
The device name is not always the last column.  Each kind (TRAFFIC, THREAT, etc.) is different.

rfaircloth-splunk:
> Check
> https://docs.paloaltonetworks.com/pan-os/7-1/pan-os-admin/monitoring/use-syslog-for-monitoring/syslog-field-descriptions
> The device name is not always the last column. Each kind (TRAFFIC, THREAT, etc.) is different.

Correct I only used the fields until the dvc_name field there are other fields after no reason to parse them and use memory.

Issue #396, opened 2020-04-06
Support Changing Diskbuffer Directory


w3ttr3y:
This shouldn't matter when people are using the container, but for people who have to bring their own environment, they are typically running in fairly restrictive environments.  Those restrictions may mean that they are only allowed to write files to certain directories or that /var/lib isn't large enough to support the desired disk buffering size and they aren't able to get /var/lib resized, etc.

My suggestion would be to add another environmental variable (say SC4S_DEST_SPLUNK_HEC_DISKBUFF_DIR) which when set would change the output directory.

There is a caveat in the notes about the dir() setting on disk buffers that says that if there is a persist file, it includes the directory that was previously used so syslog-ng will fail to use the correct setting.  While that's unfortunate, in a new setup that wouldn't be an issue and I think we should favor having options.

PR #397, opened 2020-04-06
Update splunk_hec.conf.tmpl

w3ttr3y:
Added dir argument to disk buffer in splunk_hec_conf.tmpl fixing #396 

This change adds code to look for the SC4S_DEST_SPLUNK_HEC_DISKBUFF_DIR environmental variable and if present, it uses that directory for the directory of the disk buffer.

The dir() option allows you to change the directory used for disk buffers.  For 3.16 the feature is document at
https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.16/administration-guide/disk-buffer

The caveat is if you have already encountered disk buffering, there will be a persist file in the previous location (default /var/lib/syslog-ng) and that persist file will prevent syslog-ng from sucessfully using the new location.


w3ttr3y:
Added documentation for the variable as requested.

PR #422, opened 2020-04-20
Add f5 bigip irule and ASM filters

mchavda-splunk:


package/etc/context_templates/splunk_index.csv.example
```diff
@@ -27,6 +27,7 @@
 #forcepoint_webprotect,index,netproxy
 #f5_bigip,index,netops
 #f5_bigip_irule,index,netops
+#f5_bigip_asm,index,netops
```
rfaircloth-splunk, line 30: index should be netwaf if we don't have this index defined add it to the indexes.conf and docs

mchavda-splunk: Updated index from netops to netwaf for the f5_bigip_asm events

mchavda-splunk: the netwaf index is already defined in the indexes.conf

PR #423, opened 2020-04-20
Remove soup support for zscaler LSS

mbonsack:
* Enforce message/netmask globbing for zscaler LSS.  Hostname globbing not possible because `$HOST` has not been parsed yet.


mbonsack:
Replaced with PR 426

PR #449, opened 2020-05-11
Adding the support for Cisco WSA

nkaleiya:


mbonsack:
In `vendor_product_by_source.csv.example`, add the entries for cisco_wsa* in the proper place (in roughly alphabetical order); e.g. after `f_cisco_meraki`, rather than at the end.

nkaleiya:
> In `vendor_product_by_source.csv.example`, add the entries for cisco_wsa* in the proper place (in roughly alphabetical order); e.g. after `f_cisco_meraki`, rather than at the end.

Got it! I have moved it and updated the PR.

PR #478, opened 2020-05-20
Mchavda splunk/imperva waf test cases 424

rfaircloth-splunk:


mchavda-splunk:
@rfaircloth-splunk last commit [091053b]([https://github.com/splunk/splunk-connect-for-syslog/pull/478/commits/091053b1ff354f11c068e7d55111fb46275dfd0f) might require some modification. your help would be much appreciated

PR #494, opened 2020-05-28
Dyanamic reverse DNS

rfaircloth-splunk:
When the host is empty or an IP try reverse DNS


tmoser79:
I suggest that you rename Dyanamic to Dynamic :-)

PR #514, opened 2020-06-11
Extract Source based on the product name & Update Doc

mchavda-splunk:


rfaircloth-splunk:
So that future product names can be automatically supported could we instead of using a constant extract with regex the value of the field used in the filter replace space with _ and force lowercase?

mchavda-splunk:
@rfaircloth-splunk I'm not able to find any example to convert the macro value into the lowercase. Could you please provide some doc link for the reference.

mchavda-splunk:
In the Splunk search, the values are case insensitive. Do you think we should still convert it into the lowercase?

rfaircloth-splunk:
the reason I would normalize case is that search field=value is case insensitive, props/source:: license metrics, internal stats, and | stats | tstats are not. if there are cases where the vendor is inconsistent it would make use more difficult.

mchavda-splunk:
Gotcha. updated PR.  please review it.

PR #552, opened 2020-07-03
Fixed the issue of CISE_Alarm events for Cisco ISE

jashah-splunk:
CISE_Alarm events of Cisco ISE, Do not contain `ISE.num`, `ISE.seq`, etc fields and the format is different than other CISE_* fields. Hence added a check that if the event is CISE_Alarm then index the event as is and do not wait to group it with others.


jashah-splunk:
> Please add a test case for this message type

Test case added.

PR #559, opened 2020-07-09
Quickstart_guide.md added

nandinivij:
Added quickstart guide with just implementation steps


nandinivij:
> Two things we need to add this to the menu and we need to specify which OS this works for I think its RHEL8

This works with RHEL 7.6(podman) and Amazon Linux(docker)

PR #576, opened 2020-07-23
[Filteradd] syslog-ng loggen filter

nandinivij:


mbonsack:
I believe the default for loggen is _not_ to use RFC 5424 format.  We will need to carefully articulate in the docs how we want customers to invoke `loggen` if we are going to limit the log path to `rfc5424_strict` events.  Otherwise, the log path should account for legacy BSD syslog as well as RFC 5424.

PR #630, opened 2020-08-11
Update faq.md

mkarlstrand-splunk:
HA Q&A to FAQ


mbonsack:
Do you want to remove the question immediately following?  It asks the same thing...

mkarlstrand-splunk:
I wasn't sure if it was exactly the same question. If they are duplicate then, yes remove it. thx

PR #635, opened 2020-08-14
ADDON-28629: Added filters for all sourcetype

tbalar-splunk:
JIRA Ticket: https://jira.splunk.com/browse/ADDON-28629


nandinivij:
Hi @tbalar-splunk, Please check the build test is failing for added test. 
https://app.circleci.com/pipelines/github/splunk/splunk-connect-for-syslog/2972/workflows/784b10ce-05fe-4510-a8ce-5f528cf5475d/jobs/12012/steps
Thank you

package/etc/conf.d/filters/cisco/esa.conf
```diff
@@ -0,0 +1,44 @@
+filter f_cisco_esa {
+    match("cisco_esa_12_5", value("fields.sc4s_vendor_product") type(glob))
```
rfaircloth-splunk, line 2: if we can avoid it we don't want this to be product-version specific because admins upgrade stuff and don't tell the log admins

package/etc/conf.d/filters/cisco/esa.conf
```diff
@@ -0,0 +1,44 @@
...
+filter f_cisco_esa_http {
+    match('https?\s*service\s*', value("MESSAGE"))
+    or match('[sS]ession\s*\w+\s*from\s*', value("MESSAGE"))
+    or match('\s*PERIODIC\s*REPORTS\s*:\s*', value("MESSAGE"))
+    or match('\s*req\s*:.+\s*user\s*:\s*.+\s*id\s*:\s*', value("MESSAGE"))
+    or match('testmaillog\s*:\s*[iI]nfo\s*:', value("MESSAGE"))
+    or match('[sS]ystem\s*is\s*coming\s*up', value("MESSAGE"))
+};
+
+filter f_cisco_esa_cef {
```
rfaircloth-splunk, line 14: Cef formats should not be custom filters we can use the common log path for that

package/etc/conf.d/filters/cisco/esa.conf
tbalar-splunk: Here we want to differentiate the events based on the host and so we have assigned the sc4s_vendor_product as cisco_esa_12_5. Hence all the logs coming from that host will be assigned to this vendor_product. Do you suggest we change that? and If yes, What is the condition that we put?

package/etc/conf.d/filters/cisco/esa.conf
tbalar-splunk: Okay, Removed the **f_cisco_esa_cef** filter

tbalar-splunk:
@rfaircloth-splunk  With the new code changes in the develop branch, the existing test cases are failing.
Hence, declining this PR.

PR #636, opened 2020-08-14
Update CEF Parser

mbonsack:
* Update CEF parser to use standard 3164 parsing
* [filtermod] [breaking] Change indexed field `sc4s_cef_version` to `cef_version` to align with other CEF indexed fields
* [filtermod] [breaking] Change value of `sc4s_syslog_format` from `cef` to `CEF` (it is an acronym)
* Fix minor but in Imperva WAF test


package/etc/conf.d/conflib/_common/templates.conf
```diff
@@ -39,7 +39,7 @@ template t_everything {
 # ===============================================================================================
 
 template t_cef_hdr_msg {
-        template("${3}");
+        template("CEF:${MESSAGE}");
```
rfaircloth-splunk, line 42: I don't think we need the CEF: really

rfaircloth-splunk: Modified on PR to remove CEF: and trim white space

PR #695, opened 2020-09-07
Addon 29030 update filter to collect data in f5 syslog

dhruvinshah-crest:


package/etc/conf.d/log_paths/lp-f5_bigip.conf.tmpl
```diff
@@ -14,7 +14,13 @@ filter f_f5_bigip {
     or program("tmm\d?")
     or program('^f5_irule=')
     or program('F5')
-    or message('^f5_asm=Splunk-F5-ASM');
+    or message('^f5_asm=Splunk-F5-ASM')
```
rfaircloth-splunk, line 17: the following lines don't reliably indicate the event is F5 as the terms could be used by other products. We must relay on the host match to set vendor_product then classify inside the log path

rfaircloth-splunk (changes requested):
see comments

package/etc/conf.d/log_paths/lp-f5_bigip.conf.tmpl
nkaleiya: @rfaircloth-splunk, As per our Splunk doc about the irules configuration, it is mentioned that f5_asm should be equal to Splunk-F5-ASM. Reference: https://docs.splunk.com/Documentation/AddOns/released/F5BIGIP/Setup

rfaircloth-splunk: Line 17 wasn't changed the comment applies to the new lines below

dhruvinshah-crest: Yes true, Understand your point and removed all those matches as we already have a match for the host on the fist line as `match("^f5_bigip", value("fields.sc4s_vendor_product"))`.

rfaircloth-splunk (approved):

PR #746, opened 2020-10-02
Update test_avi_vantage.py

rfaircloth-splunk:


nandinivij:
Please share why do we add host at the end of the test data?

mbonsack (approved):

PR #856, opened 2020-11-20
Feature/addon 31021 create and update juniper filters

ssamaddar-crest:
ADDON-31021: Updated the filters to support the following message tags:
- PFE_FW_SYSLOG_ETH_IP
- SNMP_TRAP_LINK_UP
- SNMP_TRAP_LINK_DOWN
- ESWD_STP_STATE_CHANGE_INFO
- ESWD_DAI_FAILED

### DIFF
--- a/docs/sources/Juniper/index.md
+++ b/docs/sources/Juniper/index.md
@@ -17,6 +17,7 @@
 | juniper:junos:idp:structured   | None                                                             |
 | juniper:junos:aamw:structured   | None                                                             |
 | juniper:junos:secintel:structured   | None                                                             |
+| juniper:junos:snmp   | None                                                             |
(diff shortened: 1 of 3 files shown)


rfaircloth-splunk (changes requested):
juniper:junos:snmp isn't or should not be a new sourcetype this could be a good eventtype

rfaircloth-splunk (changes requested):
I see the reason for the new sourcetype, but the implementation is incomplete the following additional syslog tags should be included
https://www.juniper.net/documentation/en_US/junos/topics/concept/standard-snmp-traps-overview.html

rfaircloth-splunk (approved):

PR #880, opened 2020-12-16
feat(palo): Support Traps4 CEF Format

rfaircloth-splunk:


nandinivij (approved):

mbonsack:
What is this index? `Palo Alto Networks_Traps Agent,index,epintelexit`?
Should it be just `epintel`?

PR #987, opened 2021-02-22
ADDON-26751 : Added sc4s filter for the Splunk Add-on for CyberArk

hsoladhra-crest:
Added sc4s filter for the Splunk Add-on for CyberArk
Jira Ref: [ADDON-26751](https://jira.splunk.com/browse/ADDON-26751)


hsoladhra-crest:
Declining this PR as CyberArk events are CEF format So do not need to add filter for it.

PR #1032, opened 2021-03-16
feat(f5-bigip): Added new support for sourcetypes

darshm-crest:
Added support for the following sourcetypes.

- f5:bigip:secure
- f5:bigip:ltm:ssl:error
- f5:bigip:ltm:tcl:error
- f5:bigip:ltm:traffic
- f5:bigip:ltm:log:error


package/etc/conf.d/conflib/syslog/app-f5_bigip.conf
```diff
@@ -27,6 +27,47 @@ block parser f5_bigip-parser() {
                 set("f5_bigip_nix", value("fields.sc4s_vendor_product"));
                 set("nix:syslog", value(".splunk.sourcetype"));
             };
+        } elif {
+            filter{
+                message('[0-9A-Fa-f]{4}\d+:\d{1,2}:\sTCL\serror:\s.+\/(?:\s[^\s]+.*ERR_NOT_SUPPORTED)?.*');
```
rfaircloth-splunk, line 32: regex on message is expensive, in these it looks like this only applies if program is tmm use an outer if and inside that if do if else logic for the message rules to reduce the cpu cost

rfaircloth-splunk (changes requested):

rfaircloth-splunk:
closed this PR abandoned, I made similar changes in another PR with customer direct feedback

PR #1146, opened 2021-05-05
doc(microfocus): Update doc links for microfocus

### LINKED ISSUE #895: Readthedocs: Microfocus Arcsight: Product reference are incorrect
Here is the link to the webpage:
[https://splunk-connect-for-syslog.readthedocs.io/en/master/sources/Microfocus/](https://splunk-connect-for-syslog.readthedocs.io/en/master/sources/Microfocus/)

1) The referenced product manual is that of Imperva and not Arcsight.

2) It's not clear from the webpage what logs it's meant to collect from Arcsight, such as those sent via the Forwarder Collector or specific internal logs from Arcsight? It would be useful to have this cleared up a bit if the referenced documentation isn't explicit.

3) Also, any reason why the CEF TA isn't published on Splunkbase? There is one in Splunkbase called CEF Extraction Add-on for Splunk but perhaps the one being referenced is better maintained?

rfaircloth-splunk:
Fixes #895

### DIFF
--- a/.github/workflows/coc.yaml
+++ b/.github/workflows/coc.yaml
@@ -26,6 +26,7 @@ jobs:
           #below are the optional inputs - If the optional inputs are not given, then default values will be taken
           remote-organization-name: splunk
           remote-repository-name: cla-agreement
+          custom-pr-sign-comment: "I have read the Code of Conduct and I hereby accept the Terms"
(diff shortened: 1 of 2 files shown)


rfaircloth-splunk:
recheck

rfaircloth-splunk:
recheck

PR #1357, opened 2021-12-08
feat: Support Symantec DLP

hsoladhra-crest:


package/etc/conf.d/conflib/raw/app-symantec_dlp.conf
```diff
@@ -0,0 +1,18 @@
...
+                sourcetype('symantec:dlp:syslog')
+                vendor_product("symantec_dlp")
+            );              
+        };          
+
+   };
+};
+application symantec_dlp[sc4s-syslog] {
+	filter { 
+        message('incident_id' type(string) flags(substring));
```
rfaircloth-splunk, line 15: I can imagine the string "incident_id" could appear in other syslog source this isn't safe enough for a parser we need something more unique or the filter will need to be host/ip based

rfaircloth-splunk (changes requested):

package/etc/conf.d/conflib/raw/app-symantec_dlp.conf
hsoladhra-crest: @rfaircloth-splunk Updated as per the review comments

rfaircloth-splunk (changes requested):
The parser is in the wrong folder move to the same as the other similar parsers

rfaircloth-splunk (approved):

PR #1682, opened 2022-05-03
Update switch.md

roberttoups:
The file path for app parsers uses an underscore and not a dash in the current version.

`[root@nyc128-sc4s-01 app_parsers]# pwd
/opt/sc4s/local/config/app_parsers`


rjha-splunk:
Please use conventional commit to raise the PR, thank you for contribution.

Issue #1754, opened 2022-07-18
v2.32.3 - Unable to start container when Syslog Standard destination set in ENV


davidbattyJDS:
The following error is seen on startup:

```
Jul 18 08:57:10 sc4s systemd[1]: Starting SC4S Container...
Jul 18 08:57:10 sc4s systemd[1]: Started SC4S Container.
Jul 18 08:57:11 sc4s docker[3772]: SC4S_ENV_CHECK_HEC: Splunk HEC connection test successful to index=main for sourcetype=sc4s:fallback...
Jul 18 08:57:11 sc4s docker[3772]: SC4S_ENV_CHECK_HEC: Splunk HEC connection test successful to index=prd_monitoring for sourcetype=sc4s:events...
Jul 18 08:57:12 sc4s docker[3772]: Traceback (most recent call last):
Jul 18 08:57:12 sc4s docker[3772]:   File "/etc/syslog-ng/conf.d/destinations/dest_syslog/plugin.py", line 76, in <module>
Jul 18 08:57:12 sc4s docker[3772]:     transport=transport,
Jul 18 08:57:12 sc4s docker[3772]: NameError: name 'transport' is not defined
Jul 18 08:57:12 sc4s docker[3772]: [2022-07-17T22:57:12.278952] confgen: Generator program returned with non-zero exit code; reference='/etc/syslog-ng/conf.d/destinations/dest_syslog/plugin.conf:3:1', context='root', block='dest_syslog', exec='/etc/syslog-ng/conf.d/destinations/dest_syslog/plugin.py', rc='256'
Jul 18 08:57:12 sc4s docker[3772]: Error parsing config, syntax error, unexpected LL_ERROR, expecting end of file in /etc/syslog-ng/conf.d/destinations/dest_syslog/plugin.conf:3:1-3:12:
Jul 18 08:57:12 sc4s docker[3772]: Included from /etc/syslog-ng/syslog-ng.conf:49:1-49:1:
Jul 18 08:57:12 sc4s docker[3772]: 44      @include "conf.d/local/config/sources/*.conf"
Jul 18 08:57:12 sc4s docker[3772]: 45      @include "conf.d/local/config/sources/*/*.conf"
Jul 18 08:57:12 sc4s docker[3772]: 46
Jul 18 08:57:12 sc4s docker[3772]: 47      @include "conf.d/destinations/*.conf"
[...]

PR #1758, opened 2022-07-20
Additional fix for #1754

pwickham:
We noticed that SC4S still would not start as the tls value was not being passed to the jinja template and so the required tls options were not being set.


pwickham:
I have read the Code of Conduct and I hereby accept the Terms

rjha-splunk (approved):

PR #1759, opened 2022-07-20
[FIX] fix netapp log parser by adding capital letters option in host …

mateuszpierzchala-splunk:
[FIX] fix netapp log parser by adding capital letters option in host field


mateuszpierzchala-splunk:
Closes #1759

PR #1948, opened 2022-12-05
vscode: add static file association for syslog-ng conf files

zyphermonkey:


ajasnosz:
VSCode should not be added to main repository code.

PR #1991, opened 2023-01-31
feat: juniper doc updates and liveaction parser

bparmar-splunk:
- Liveaction Parser
- Juniper default index doc update


package/etc/conf.d/conflib/syslog/app-syslog-liveaction_livenx.conf
```diff
@@ -0,0 +1,18 @@
...
+            r_set_splunk_dest_default(
+                index("netops")
+                sourcetype('liveaction:livenx')
+                vendor("liveaction")
+                product("livenx")
+            )
+        };
+    };
+};
+application app-syslog-liveaction_livenx[sc4s-syslog] {
```
rjha-splunk, line 13: looks like liveaction is a PROGRAM , can you check sc4s-syslog-pgm topic here, we have to filter based on program macro.

package/etc/conf.d/conflib/syslog/app-syslog-liveaction_livenx.conf
```diff
@@ -0,0 +1,18 @@
...
+    filter {
+        message('%LIVEACTION'  type(string)  flags(prefix));
```
rjha-splunk, line 15: program('<string to use>' type(string) flags(prefix))

bparmar-splunk: Message is not considering `%LIVEACTION` as program and hence I have parsed it as message.

rjha-splunk (approved):

PR #2033, opened 2023-03-29
feat: log_iw_size added for UDP

rjha-splunk:


rjha-splunk:
It will expose log-iw-size for UDP

PR #2117, opened 2023-07-21
fix: ca and tls certificate based on alpine based container

rjha-splunk:
It will fix the CA and TLS certs , as we have changed the base image from RHEL to Alpine , the existing way of updating the certificate was not working.


rjha-splunk:
Closes #2117

mstopa-splunk (approved):

PR #2131, opened 2023-08-07
fix: exclude filtering by generic program for syslog

ikheifets-splunk:


package/etc/conf.d/conflib/syslog/app-syslog-vmware_vsphere.conf
```diff
@@ -20,7 +20,6 @@ filter syslog-vmware_vsphere-esx-pgm{
     or program("esxupdate", type(string) flags(ignore-case,prefix))
     or program("healthd", type(string) flags(ignore-case,prefix))
     or program("ConfigStore", type(string) flags(ignore-case,prefix))
-    or program("crond", type(string) flags(ignore-case,prefix))
```
ikheifets-splunk, line 23: here we have filtering by program `crond` of regular syslog message. `crond` it's a very popular tool (much popular only `sshd`) by this reason it's overriding another filters.

If the customer real wanna get `sourcetype=vmware:esxlog:crond` then we can propose to use `SC4S_LISTEN_VMWARE_VSPHERE_TCP_PORT` and it will go to [this](https://github.com/splunk/splunk-connect-for-syslog/blob/main/package/etc/conf.d/conflib/netsource/app-netsource-vmware_vsphere.conf#L23) net source filter

rjha-splunk (approved):

mstopa-splunk (approved):

PR #2153, opened 2023-08-31
feat: add syslog format for a10networks

ikheifets-splunk:


package/etc/conf.d/conflib/syslog/app-syslog-a10.conf
```diff
@@ -0,0 +1,20 @@
...
+                vendor('a10networks')
+                product('vThunder')
+            );
+        };
+    };
+};
+
+application app-syslog-a10[sc4s-syslog-pgm] {
+    filter {
+        program('a10logd' type(string) flags(prefix));
```
ikheifets-splunk, line 17: @rjha-splunk @mstopa-splunk Is it okay or `program('a10logd' type(string) flags(prefix)) and message('[audit log]'  type(string)  flags(prefix))` would be better? I just really worry about that `a10logd` might be suite for different devices for that vendor

docs/sources/vendor/a10networks/vthunder.md
```diff
@@ -16,9 +16,10 @@
 
 ## Sourcetypes
 
-| sourcetype               | notes                                                                                         |
-|--------------------------|-----------------------------------------------------------------------------------------------|
-| a10networks:vThunder:cef | None                                                                                          |
+| sourcetype                  | notes                                                                                      |
+|-----------------------------|--------------------------------------------------------------------------------------------|
+| a10networks:vThunder:cef    | CEF                                                                                        |
+| a10networks:vThunder:syslog | Syslog                                                                                     |
```
ikheifets-splunk, line 22: Before that customer sent us pcap only with CEF logs, it's seems that device producing logs in 2 formats

package/etc/conf.d/conflib/syslog/app-syslog-a10.conf
mstopa-splunk: `a10logd` will be sufficient

mstopa-splunk (approved):

Issue #2180, opened 2023-09-26
app-syslog-dell_switch_n catching non powerswitch events


jrehm-mmm:
I have seen this with several different vendor products.  Most recently found with Aruba Clearpass, but have seen it happen with vmware_vsphere,isc_bind, and cisco_ios.  If an event makes it to app-syslog-dell_switch_n[sc4s-syslog] AND is from a program that contains a capital letter, it will truncted digits from the end of the hostname and prefix it to the message.

Per Ryan F - "I think the parser design is wrong in hindsight and needs to change to vps instead of syslog"

**What is the sc4s version ?**
tried with 2.23 and 3.4.3

**Is there a pcap available?**
No - I have not been able to capture a faulty message.  I can get millions of events without any getting caught by app-syslog-dell_switch_n[sc4s-syslog].  

**Example event** 
<135>2023-09-27 06:04:24,445 10.1.2.3 CPPM_System_Event 6 1 0 event_source=TACACS+ Server,level=ERROR,category=Request,description=TACACS+ request from unknown NAD=10.198.2.200,action_key=Failed,timestamp=2023-09-24 12:20:28.402-05

**dns lookup result** 
10.1.2.3=myhost01

**Splunk Event**
host=myhost (missing the 01)
_raw=01 CPPM_System_Event 6 1 0 event_source=TACACS+ Server,level=ERROR,category=Request,description=TACACS+ request from unknown NAD=10.198.2.200,action_key=Failed,timestamp=2023-09-24 12:20:28.402-05

**sc4s_tags on the above faulty event**
[wireformat:rfc|wireformat:rfc3164|source_identified|wireformat:rfc3164_isodate|.app.app-almost-syslogz-isodate|.app.app-syslog-dell_switch_n|ns_vendor:aruba|ns_product:clearpass|.source.s_ARUBA_CLEARPASS|.app.app-aruba_clearpass-post|vps]

**sc4s_tags on an event from the same sc4s_container from the same host to for the same CPPM_System_Event type**
[wireformat:rfc|wireformat:rfc3164|source_identified|wireformat:rfc3164_isodate|.app.app-almost-syslogz-isodate|ns_vendor:aruba|ns_product:clearpass|.source.s_ARUBA_CLEARPASS|.app.app-netsource-aruba_clearpass|.app.app-aruba_clearpass-post|vps]

mstopa-splunk:
[...]

PR #2235, opened 2023-10-11
fix: fix Dell EMC Powerswitch n parser

mstopa-splunk:
solves https://github.com/splunk/splunk-connect-for-syslog/issues/2180

This Pull Request introduces a breaking change to the codebase that addresses an issue related to parsing logs from Dell EMC PowerSwitch N devices. Previously, the logs were parsed based on message content using the syslog parser located at `package/etc/conf.d/conflib/syslog/app-syslog-dell_switch_n.conf`.

## Background
In issue #2180, it was reported that the existing parser had a bug and incorrectly classified events from various devices such as VMware vSphere, ISC BIND, Cisco IOS, and Aruba ClearPass. The root cause of the issue was identified as a flawed design of the parser, which did not rely on unique message characteristics of Dell PowerSwitch N devices. Instead, it improperly classified events based on the hyphen in the hostname, whenever the program name started with an uppercase letter.

## Breaking Change Details
To rectify this issue and improve the accuracy of log parsing, a breaking change was introduced. The syslog parser `package/etc/conf.d/conflib/syslog/app-syslog-dell_switch_n.conf` has been replaced with a new netsource parser located at `package/etc/conf.d/conflib/netsource/app-netsource-dell_switch_n.conf`.

## Impact
This breaking change will affect any existing functionality or configurations that relied on the previous syslog parser for Dell PowerSwitch N logs.


tests/test_dell_emc_networking.py
```diff
@@ -16,31 +16,42 @@
...
-    "{{ mark }} {{ bsd }} {{ host }}-1 TRAPMGR[trapTask]: traputil.c(721) 29331 %% NOTE 'startup-config' has changed.",
-    "{{ mark }} {{ bsd }} {{ host }}-1 UNITMGR[emWeb]: unitmgr.c(6905) 29330 %% INFO Configuration propagation successful for config type 0",
+# <189> Oct 21 09:10:54 test-dell-switch-n-1 CMDLOGGER[emWeb]: cmd_logger_api.c(83) 29333 %% NOTE CLI:10.1.3.211:administrator:User  logged in
+# <189> Oct 21 09:10:20 test-dell-switch-n-2 TRAPMGR[trapTask]: traputil.c(721) 29331 %% NOTE 'startup-config' has changed.
+# <190> Oct 21 09:10:20 test-dell-switch-n-3 UNITMGR[emWeb]: unitmgr.c(6905) 29330 %% INFO Configuration propagation successful for config type 0
+
+
+test_cases = [
+    {
+        "number": 1,
```
ikheifets-splunk, line 26: I think instead of `number` we can use in test index in array :)

tests/test_dell_emc_networking.py
```diff
@@ -16,31 +16,42 @@
...
 
 
-@pytest.mark.parametrize("event", testdata_admin)
+@pytest.mark.parametrize("case", test_cases)
 def test_dell_emc_powerswitch_nseries(
-    record_property,  get_host_key, setup_splunk, setup_sc4s, event
+    record_property, setup_splunk, setup_sc4s, case
 ):
-    host = "" + get_host_key
+    host = f'test-dell-switch-n-{case["number"]}'
```
ikheifets-splunk, line 46: I think instead of `number` we can use in test index in array or just generate this number randomly :)

package/etc/conf.d/conflib/syslog/app-syslog-dell_switch_n.conf
```diff
@@ -1,33 +0,0 @@
-block parser app-syslog-dell_switch_n() {
```
ikheifets-splunk, line 1: I have only one concern about that. We migrating from `syslog`  filter to `netsource`. I think it might break backward compatibility  for user who will update and don't configure it on custom port / sc4s-vps filter.

tests/test_dell_emc_networking.py
mstopa-splunk: you're right, done

PR #2267, opened 2023-11-17
Bump version of sc4s in ansible playbooks

ikheifets-splunk:


ansible/resources/docker_sc4s.service
```diff
@@ -8,7 +8,7 @@ Requires=docker.service
 WantedBy=multi-user.target
 
 [Service]
-Environment="SC4S_IMAGE=ghcr.io/splunk/splunk-connect-for-syslog/container2:2"
+Environment="SC4S_IMAGE=ghcr.io/splunk/splunk-connect-for-syslog/container3:latest"
```
ikheifets-splunk, line 11: Probably we need use "playbook variables" for that and substitude version using Jinja, I think it should be configurable.

mstopa-splunk: good point

mstopa-splunk (approved):

PR #2272, opened 2023-11-27
feat: enable IPv6 caching

mstopa-splunk:


package/etc/pylib/parser_source_cache.py
```diff
@@ -19,11 +19,33 @@ class LogDestination:
 
 
 def ip2int(addr):
-    return struct.unpack("!I", socket.inet_aton(addr))[0]
+    ip4_to_int = lambda addr: struct.unpack("!I", socket.inet_aton(addr))[0]
+    
+    def ip6_to_int(addr):
```
ikheifets-splunk, line 24: Can we please write a unit test on that?

package/etc/pylib/parser_source_cache.py
```diff
@@ -19,11 +19,33 @@ class LogDestination:
...
+    except OSError:
+        return ip6_to_int(addr)
 
 
 def int2ip(addr):
-    return socket.inet_ntoa(struct.pack("!I", addr))
+    int_to_ip4 = lambda addr: socket.inet_ntoa(struct.pack("!I", addr))
+
+    def int_to_ip6(num):
+        a = (num >> 64) & 0xFFFFFFFFFFFFFFFF
```
ikheifets-splunk, line 39: Can we please write a unit test on that?

.github/workflows/ci-main.yaml
```diff
@@ -227,7 +227,113 @@ jobs:
             --sc4s_host=sc4s \
             --junitxml=test-results/test.xml \
             -n 14 \
-            -k 'not lite'
+            -k 'not lite and not name_cache'
+  
+  test-ipv4-name-cache:
```
ikheifets-splunk, line 232: Probably we need to add it as [dependency for release](https://github.com/splunk/splunk-connect-for-syslog/blob/a73020bf0bbae984cf3aa229267d9366a102f723/.github/workflows/ci-main.yaml#L362), because when it fail release anyway will be done

PR #2280, opened 2023-12-05
fix: sc4s archiving

ikheifets-splunk:


package/etc/conf.d/sc4slib/app-lp-global-archive/plugin.py
```diff
@@ -28,7 +28,7 @@
     ]:
 
         msg = tm.render(
-            key=r,
+            key=r.lower(),
```
ikheifets-splunk, line 31: we have [here](https://github.com/splunk/splunk-connect-for-syslog/blob/2f21ff9028bc44ad0b3d0dc081a066994bdd9aa3/package/etc/conf.d/sc4slib/app-lp-global-archive/plugin.jinja#L3):

```'{{ key }}' eq "${fields.sc4s_vendor}_${fields.sc4s_product}"```

And it seems that we extracting from `SC4S_DEST_PAN_PANOS_ARCHIVE` the  `PAN_PANOS` and compare with `pan_panos`

ikheifets-splunk: @rjha-splunk @mstopa-splunk probably something like that would be much better:

```match('{{ key }}' value('${fields.sc4s_vendor}_${fields.sc4s_product}') flags(ignore-case))```

than this approach

mstopa-splunk (approved):

rjha-splunk (approved):

package/etc/conf.d/sc4slib/app-lp-global-archive/plugin.py
rjha-splunk: This is great idea.

PR #2291, opened 2023-12-12
docs: update metrics index

mstopa-splunk:


ikheifets-splunk (approved):

docs/resources/indexes.conf
```diff
@@ -22,11 +16,11 @@ homePath   = $SPLUNK_DB/epintel/db
 coldPath   = $SPLUNK_DB/epintel/colddb
 thawedPath = $SPLUNK_DB/epintel/thaweddb
 
-[em_metrics]
```
ikheifets-splunk, line 25: Can you please also fix [here](https://github.com/splunk/splunk-connect-for-syslog/blob/main/.indexes.conf#L65).

Issue #2347, opened 2024-02-15
Make the bluecoat sourcetype match the sourcetype defined in the Splunk bluecoat addon


myriadic:
**What is the sc4s version ?**
most recent

**Is there a pcap available?**
no

**What the vendor name?**
Symantec

**What's the product name?**
BluecoatProxy

** Feature Request description: **
Make the bluecoat sourcetype match the sourcetype defined in the Splunk bluecoat addon

** Should it support TCP or UDP?**
UDP

** Do you want to have it for local usage or prepare a github PR? **
github PR

The Bluecoat addon has "KV_MODE = none" for the bluecoat sourcetype, which means the field extractions will have to match the SC4S format.

The issue with this is that SC4S strips out the timestamp. For the field extractions to work, even after making the sourcetypes match, the timestamps in the bluecoat addon need to be marked as optional in the field extraction REGEX

mstopa-splunk:
Hi @myriadic thanks for catching this. 

sourcetype fix should be merged in one of the next releases: https://github.com/splunk/splunk-connect-for-syslog/pull/2370
In the meantime I will pass the info about incorrect REGEX to the addon owners. Can you send me an example of the event to test? The one we have in repo is not super useful for this.

mstopa-splunk:
sourcetype fix released in v3.25.0, regex needs to be fixed on the add-on side, please send a few samples to mstopa@splunk.com or to Splunk support

PR #2370, opened 2024-03-15
fix: update Bluecoat sourcetype to match TA 3.8.1

mstopa-splunk:
solves https://github.com/splunk/splunk-connect-for-syslog/issues/2347

### DIFF
--- a/docs/sources/vendor/Broadcom/proxy.md
+++ b/docs/sources/vendor/Broadcom/proxy.md
@@ -23,14 +23,14 @@ Broadcom products are inclusive of products formerly marketed under Symantec and
 
 | sourcetype     | notes                                                                                                   |
 |----------------|---------------------------------------------------------------------------------------------------------|
-| bluecoat:proxysg:access:kv        | Requires version TA 3.6                                                                                                    |
-| bluecoat:proxysg:syslog           | Requires version TA 3.6                                                                                                    |
+| bluecoat:proxysg:access:kv        | Requires version TA 3.8.1                                                                                                    |
+| bluecoat:proxysg:access:syslog           | Requires version TA 3.8.1                                                                                                    |
 
 ## Sourcetype and Index Configuration
 
 | key            | sourcetype     | index          | notes          |
 |----------------|----------------|----------------|----------------|
-| bluecoat_proxy      | bluecoat:proxysg:syslog       | netops          | none          |
+| bluecoat_proxy      | bluecoat:proxysg:access:syslog       | netops          | none          |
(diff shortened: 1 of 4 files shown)


docs/sources/vendor/Broadcom/proxy.md
```diff
@@ -23,14 +23,14 @@ Broadcom products are inclusive of products formerly marketed under Symantec and
 
 | sourcetype     | notes                                                                                                   |
 |----------------|---------------------------------------------------------------------------------------------------------|
-| bluecoat:proxysg:access:kv        | Requires version TA 3.6                                                                                                    |
-| bluecoat:proxysg:syslog           | Requires version TA 3.6                                                                                                    |
+| bluecoat:proxysg:access:kv        | Requires version TA 3.8.1                                                                                                    |
+| bluecoat:proxysg:access:syslog           | Requires version TA 3.8.1                                                                                                    |
```
ikheifets-splunk, line 27: Am I understood correctly that add-on maintainers broke backward compatibility (changed `sourcetype`) starting with 3.8.1 version ?

Can we talk with them to understand the reason?  Probably they will rollback this change if add-on not very popular :)

**If we need to break compatibility, probably good idea to provide some env var (`BLUECOAT_NEW_FORMAT=true`) that will determine we will assign old or new sourcetype for `bluecoat`. If will help us to prevent of breaking backward compatibility**

mstopa-splunk: that's a good question, 3.8.1 was released in September 2022 and I don't have access to older versions to check if that was changed on the TA side or there was a bug in SC4S

mstopa-splunk: especially that `bluecoat:proxysg:access:syslog` is more consistent with the rest than `bluecoat:proxysg:syslog`

ikheifets-splunk: @mstopa-splunk you are right, it's impossible to check, because they deleted previous versions :)

Only 3.8.1 available on Splunkbase :)

mstopa-splunk: @ikheifets-splunk please see how this has been done previously:

https://splunk.github.io/splunk-connect-for-syslog/main/sources/vendor/Dell/sonicwall/
```
Note:[¶](https://splunk.github.io/splunk-connect-for-syslog/main/sources/vendor/Dell/sonicwall/#note)
The sourcetype has been changed in version 2.35.0 making it compliant with corresponding TA.
```

the practice seems to be updating the sourcetype and adding the note

PR #2373, opened 2024-03-19
ci: scan docker image on cves

ikheifets-splunk:


.github/workflows/ci-lite.yaml
```diff
@@ -154,6 +154,21 @@ jobs:
           cache-from: type=registry,ref=${{ needs.meta.outputs.container_base }}
           cache-to: type=inline
 
+  scan-docker-image-cves:
```
ikheifets-splunk, line 157: don't requiring it as a dependency for CI/CD release, because for hotfixes it shouldn't block release

rjha-splunk (approved):

Issue #2381, opened 2024-03-28
Truncated IP Address in host field


ehlo550:
**What is the sc4s version ?**
3.22.1

**Is the issue related to the environment of the customer or Software related issue?**
software related

**Describe the bug**
After upgrading to 3.22.1 i saw that the host value got truncated for several hosts. 
Instead of 10.1.1.1 & 10.1.1.2 it was only 10
Instead of 172.20.20.1 & 172.20.30.1 it was only 172. 

I have SC4S_USE_REVERSE_DNS = true and SC4S_REVERSE_DNS_KEEP_FQDN = false
and if there is no reverse DNS entry instead of the original IP address only a truncated version is used in the HOST field. 

**To Reproduce**
1. SC4S_USE_REVERSE_DNS = true
2. SC4S_REVERSE_DNS_KEEP_FQDN = false
3. Ingest logs from a Device where no hostname can be extracted from the message and where no DNS reverse entry is available
4. See wrong host in splunk

mstopa-splunk:
hi @ehlo550 can you send a raw event or point me to the problematic vendor/product? I failed to reproduce with a random message:
![image](https://github.com/splunk/splunk-connect-for-syslog/assets/139441697/b92509c2-7158-45f1-a614-8d0010995f37)

ehlo550:
Hi @mstopa-splunk,

i really do not think this is vendor specific:
![image](https://github.com/splunk/splunk-connect-for-syslog/assets/79267170/c6f734b3-92e7-4fc9-a637-df4e288f3f1f)

![image](https://github.com/splunk/splunk-connect-for-syslog/assets/79267170/e8975368-33c3-4494-9c15-ef13a204e895)

Regards

ehlo550:
I can reproduce this with
`echo '<189>409882: Mar 28 2024 16:51:54.869 CET: %DOT1X-5-FAIL: Switch 1 R0/0: sessmgrd: Authentication failed for client (0000.0000.0000) with reason (Timeout) on Interface Gi1/0/19 AuditSessionID 0A6414AC0000DD9083C9DCF0 Username: anonymous' | nc -u -w0 <sc4s_ip> 514`

mstopa-splunk:
@ehlo550 thank you, reproduced on my end

mstopa-splunk:
@ehlo550 from which SC4S version did you upgrade?

ikheifets-splunk:
[...]

PR #2383, opened 2024-04-03
fix: fix SC4S_USE_REVERSE_DNS

mstopa-splunk:
solves https://github.com/splunk/splunk-connect-for-syslog/issues/2381


tests/test_reverse_dns.py
```diff
@@ -0,0 +1,93 @@
...
+# license that can be found in the LICENSE-BSD2 file or at
+# https://opensource.org/licenses/BSD-2-Clause
+
+import pytest
+import socket
+
+from package.etc.pylib.parser_fix_dns import FixHostnameResolver, FixFQDNResolver
+
+
+class LogMessage:
```
ikheifets-splunk, line 13: Can we [import LogMessage from  syslogng](https://axoflow.com/docs/axosyslog-core/chapter-sources/python-source/python-source-logmessage/#logmessage-method-create-log-message-objects)?

ikheifets-splunk (approved):

tests/test_reverse_dns.py
ikheifets-splunk: ok, understood :)

PR #2411, opened 2024-04-18
docs: update podman-systemd-general.md

jenworthington:
incorporated Maciek's comments from previous PR

### DIFF
--- a/docs/gettingstarted/podman-systemd-general.md
+++ b/docs/gettingstarted/podman-systemd-general.md
@@ -101,7 +101,7 @@ If you do not see this, try the following before proceeding to deeper-level trou
 * Execute the following command to check the sc4s startup process running in the container.
 
 ```bash
-docker logs SC4S
+podman logs SC4S
 ```
 
 You should see events similar to those below in the output:
--- a/docs/gettingstarted/podman-systemd-general.md
+++ b/docs/gettingstarted/podman-systemd-general.md
@@ -145,7 +145,7 @@ mkdir -p /home/sc4s/tls
 podman system migrate
 ```
 
-3. Next, you need to load the new environment variables. To do this, you can temporarily switch to any other user, and then log back in as `sc4s`. When logging in as `sc4s`, avoid using the `su` command, as it won't load the new variables. Instead, you can use, for example, the command `ssh sc4s@localhost`.
+3. Next, load the new environment variables. To do this, you can temporarily switch to any other user, and then log back in as the SC4S user. When logging in as the SC4S user, don't use the 'su' command, as it won't load the new variables. Instead, you can use, for example, the command ssh sc4s@localhost.
 
 4. Create unit file in ```~/.config/systemd/user/sc4s.service``` with the following content:
 ```editorconfig
--- a/docs/gettingstarted/podman-systemd-general.md
+++ b/docs/gettingstarted/podman-systemd-general.md
@@ -1,50 +1,48 @@
 # Install podman
 
-Refer to [Installation](https://podman.io/getting-started/installation)
+See [Podman product installation docs](https://podman.io/getting-started/installation) for information about working with your Podman installation.
 
-NOTE: [READ FIRST (IPv4 forwarding)](./getting-started-runtime-configuration.md#ipv4-forwarding)
+Before performaing the tasks described in this topic, make sure you are familiar with using IPv4 forwarding with SC4S. See [IPv4 forwarding ](./getting-started-runtime-configuration.md#ipv4-forwarding).
 
 # Initial Setup
 
-* IMPORTANT:  Always use the _latest_ unit file (below) with the current release.  By default, the latest container is
-automatically downloaded at each restart.  Therefore, make it a habit to check back here regularly to be sure any changes
-that may have been made to the template unit file below (e.g. suggested mount points) are incorporated in production prior
-to relaunching via systemd.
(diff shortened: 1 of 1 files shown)


docs/gettingstarted/podman-systemd-general.md
```diff
@@ -145,7 +145,7 @@ mkdir -p /home/sc4s/tls
 podman system migrate
 ```
 
-3. Next, you need to load the new environment variables. To do this, you can temporarily switch to any other user, and then log back in as `sc4s`. When logging in as `sc4s`, avoid using the `su` command, as it won't load the new variables. Instead, you can use, for example, the command `ssh sc4s@localhost`.
+3. Next, load the new environment variables. To do this, you can temporarily switch to any other user, and then log back in as the SC4S user. When logging in as the SC4S user, don't use the 'su' command, as it won't load the new variables. Instead, you can use, for example, the command ssh sc4s@localhost.
```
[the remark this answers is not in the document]
jenworthington: Yes, thanks for calling that out, my markup might be a little rusty.

docs/gettingstarted/podman-systemd-general.md
```diff
@@ -1,50 +1,48 @@
 # Install podman
 
-Refer to [Installation](https://podman.io/getting-started/installation)
+See [Podman product installation docs](https://podman.io/getting-started/installation) for information about working with your Podman installation.
 
-NOTE: [READ FIRST (IPv4 forwarding)](./getting-started-runtime-configuration.md#ipv4-forwarding)
+Before performaing the tasks described in this topic, make sure you are familiar with using IPv4 forwarding with SC4S. See [IPv4 forwarding ](./getting-started-runtime-configuration.md#ipv4-forwarding).
```
mstopa-splunk, line 5:
```suggestion
Before performing the tasks described in this topic, make sure you are familiar with using IPv4 forwarding with SC4S. See [IPv4 forwarding ](./getting-started-runtime-configuration.md#ipv4-forwarding).
```

docs/gettingstarted/podman-systemd-general.md
```diff
@@ -1,50 +1,48 @@
...
 
 # Initial Setup
 
-* IMPORTANT:  Always use the _latest_ unit file (below) with the current release.  By default, the latest container is
-automatically downloaded at each restart.  Therefore, make it a habit to check back here regularly to be sure any changes
-that may have been made to the template unit file below (e.g. suggested mount points) are incorporated in production prior
-to relaunching via systemd.
+NOTE: Make sure to use the latest unit file, which is provided here, with the current release.  By default, the latest container is
```
mstopa-splunk, line 9:
```suggestion
NOTE: Make sure to use the latest unit file, which is provided here, with the current release. By default, the latest container is
```

docs/gettingstarted/podman-systemd-general.md
```diff
@@ -1,50 +1,48 @@
...
 ```ini
 --8<--- "docs/resources/podman/sc4s.service"
 ```
 
-* Execute the following command to create a local volume that will contain the disk buffer files in the event of a communication
-failure to the upstream destination(s).  This will also be used to keep track of the state of syslog-ng between restarts, and in
-particular the state of the disk buffer.  This is a required step.
+2. Execute the following command to create a local volume, which contains the disk buffer files in the event of a communication
+failure, to the upstream destinations.  This volume will also be used to keep track of the state of syslog-ng between restarts, and in
+particular the state of the disk buffer.
```
mstopa-splunk, line 20: can we do something like:

```
2. Execute the following command to create a local volume. The volume contains the disk buffer files and will also be used to keep track of the state of syslog-ng between restarts.
```

docs/gettingstarted/podman-systemd-general.md
```diff
@@ -82,26 +80,25 @@ sudo systemctl stop sc4s
...
-execute the following search in Splunk:
+configuration is correct.  After this step is complete, verify SC4S is properly communicating with Splunk by
+executing the following search in Splunk:
 
 ```ini
 index=* sourcetype=sc4s:events "starting up"
 ```
 
-This should yield an event similar to the following:
+This should yield an event similar to the following when the startup process proceeds normally (without syntax errors).
```
mstopa-splunk, line 90: can we cut off `(without syntax errors)`? I know it was in the doc before but other kinds of errors can disrupt the startup process as well

docs/gettingstarted/podman-systemd-general.md
```diff
@@ -82,26 +80,25 @@ sudo systemctl stop sc4s
...
+If you do not see this, try the following before proceeding to deeper-level troubleshooting:
 
 * Check to see that the URL, token, and TLS/SSL settings are correct, and that the appropriate firewall ports are open (8088 or 443).
 * Check to see that the proper indexes are created in Splunk, and that the token has access to them.
 * Ensure the proper operation of the load balancer if used.
-* Lastly, execute the following command to check the sc4s startup process running in the container.
+* Execute the following command to check the sc4s startup process running in the container.
 
 ```bash
 docker logs SC4S
```
mstopa-splunk, line 104:
```suggestion
podman logs SC4S
```

docs/gettingstarted/podman-systemd-general.md
mstopa-splunk: `(without syntax errors)` remained

PR #2420, opened 2024-04-24
docs: update docker-systemd-general.md

jenworthington:
For this statement:  "Make it a habit to check back here regularly to be sure any changes that may have been made to the template unit file below" is there a more concrete way to say this? I feel like the customer is going to be confused about how often they are expected to check back in. is it with every release?


mstopa-splunk:
> For this statement: "Make it a habit to check back here regularly to be sure any changes that may have been made to the template unit file below" is there a more concrete way to say this? I feel like the customer is going to be confused about how often they are expected to check back in. is it with every release?

@jenworthington that's a good point. In fact this file hasn't been changed for at least 8 months but I would still say that it's a good practice to check with every upgrade of SC4S on the customer side

docs/gettingstarted/docker-systemd-general.md
```diff
@@ -1,73 +1,82 @@
...
 sudo systemctl restart sc4s
 ```
 
-If changes were made to the configuration Unit file above (e.g. to configure with dedicated ports), you must first stop SC4S and re-run
+## Implement unit file changes
+If you made changes to the configuration unit file, for example to configure with dedicated ports, you must stop SC4S and re-run the systemd configuration commands to implement your changes.
+
+1. Stop SC4S:
+ ```bash
+sudo systemctl stop sc4s
```
mstopa-splunk, line 77: duplicated line 83

mstopa-splunk:
@jenworthington ready for the final pass

PR #2434, opened 2024-04-30
feat: source port validation

ikheifets-splunk:


package/sbin/source_ports_validator.py
```diff
@@ -0,0 +1,42 @@
...
+
+        if port in ["disabled", ""]:
+            continue
+        elif not is_valid_port(port):
+            logger.error(f"{env_var}: {port} should be valid int [0, 10000]")
+        elif source != "DEFAULT" and port in ["514", "614", "6514"]:
+            logger.error(f"{env_var}: Impossible to use default ports like {port}")
+        elif (port, proto) in busy_ports:
+            logger.error(f"{env_var}: {port} already used in another source, use unique port")
+        else:
```
ikheifets-splunk, line 36: @rjha-splunk @mstopa-splunk What do you think **are we also need check that port is open** ?

package/sbin/source_ports_validator.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+    busy_ports = set()
+    for source, port, proto in source_ports:
+        env_var = f"SC4S_LISTEN_{source}_{port}_PORT"
+
+        if port in ["disabled", ""]:
+            continue
+        elif not is_valid_port(port):
+            logger.error(f"{env_var}: {port} should be valid int [0, 10000]")
```
mstopa-splunk, line 31: message is like this:
```
SC4S_LISTEN_MERAKI_SWITCHES_514_PORT: Impossible to use default ports like 514
```

Can you make it more like this:
```
[source_ports_validator.py ERROR] Invalid source port detected for MERAKI_SWITCHES: 514/udp. Must be within the range [0, 1000].
[source_ports_validator.py ERROR] Invalid source port detected for MERAKI_SWITCHES: 514/udp. Default port usage not allowed (514, 614, 6514).
[source_ports_validator.py ERROR] Invalid source port detected for MERAKI_SWITCHES: 514/udp. Port number is not unique and has already been used for another source.
```

mstopa-splunk:
@ikheifets-splunk this can also be documented when approved by @rjha-splunk

package/sbin/source_ports_validator.py
ikheifets-splunk: I using `env var`, because it would be clear what you need to fix.
When you writing `MERAKI_SWITCHES` it would be unclear for users what fix

docs/troubleshooting/troubleshoot_SC4S_server.md
```diff
@@ -102,6 +102,20 @@ just _one_ bad index will "taint" the entire batch (in this case, 1000 events) a
 imperative that the container logs be free of these kinds of errors in production._ You can use the alternate HEC debug destination (below)
 to help debug this condition by sending direct "curl" commands to the HEC endpoint outside of the SC4S setting.
 
+## Invalid SC4S listen ports
+
+SC4S can exclusively grant some port to device by using environment var `SC4S_LISTEN_{vendor}_{product}_{TCP/UDP/TLS}_PORT={port}`.
```
mstopa-splunk, line 107: What do you think about updating https://splunk.github.io/splunk-connect-for-syslog/main/sources/#unique-listening-ports ?

- removing `Using the default ports as unique listening ports is discouraged since it can lead to unintended consequences. There were cases of customers using port 514 as the unique listening port dedicated for a particular vendor and then sending other events to the same port, which caused some of those events to be misclassified.`
- adding new section about the rules of validation

package/sbin/source_ports_validator.py
mstopa-splunk: but can we include the log level and logging utility name?

mstopa-splunk: discussed and decided not to include the log level and utility name

mstopa-splunk (approved):

docs/troubleshooting/troubleshoot_SC4S_server.md
```diff
@@ -102,6 +102,20 @@ just _one_ bad index will "taint" the entire batch (in this case, 1000 events) a
 imperative that the container logs be free of these kinds of errors in production._ You can use the alternate HEC debug destination (below)
 to help debug this condition by sending direct "curl" commands to the HEC endpoint outside of the SC4S setting.
 
+## Invalid SC4S listen ports
+
+[SC4S can exclusively grant some port to device by using environment var `SC4S_LISTEN_{vendor}_{product}_{TCP/UDP/TLS}_PORT={port}`](https://splunk.github.io/splunk-connect-for-syslog/main/sources/#unique-listening-ports).
```
jenworthington, line 107: Use environment variables to let SC4S exclusively grant a port to a device.

docs/troubleshooting/troubleshoot_SC4S_server.md
```diff
@@ -102,6 +102,20 @@ just _one_ bad index will "taint" the entire batch (in this case, 1000 events) a
...
+
+During startup SC4S validating that listen ports configured correctly and in case of misconfiguration will see it on logs.
```
jenworthington, line 109: During startup, SC4S validates that listening ports are configured correctly, and in case of misconfiguration, you will be able see any issues in (which log?) logs.

docs/troubleshooting/troubleshoot_SC4S_server.md
```diff
@@ -102,6 +102,20 @@ just _one_ bad index will "taint" the entire batch (in this case, 1000 events) a
...
+
+Example of error messages when listen ports for `MERAKI SWITCHES` configured incorrectly:
```
jenworthington, line 111: You may receive an error message similar to the following if listening ports for `MERAKI SWITCHES` are configured incorrectly:

PR #2496, opened 2024-06-12
doc: Updated the docker-podman-offline document with the correct image

cwadhwani-splunk:


cwadhwani-splunk:
> Please use conventional commit to raise PR , fix is used for minor version change , for docs please use doc

Done

rjha-splunk (approved):

PR #2511, opened 2024-06-25
fix: upgrade deps

ikheifets-splunk:


mstopa-splunk:
https://github.com/splunk/splunk-connect-for-syslog/pull/2588

Issue #2534, opened 2024-07-22
using PARALLELIZE feature for TCP crushes the container


olgaktoronto:
using PARALLELIZE feature for TCP crushes the container. Support is trying to replicate in the lab

**Was the issue replicated by support?**

**What is the sc4s version ?**

**Which operating system (including its version) are you using for hosting SC4S?**
Amazon EC2 c5.4xlarge/, BIOS 1.0 10/16/2017 – 32GB RAM, 64GB+600GB storage.

* Linux  5.14.0-362.24.1.el9_3.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Feb 15 07:18:13 EST 2024 x86_64 x86_64 x86_64 GNU/Linux
* Red Hat Enterprise Linux release 9.3 (Plow)

**Which runtime (Docker, Podman, Docker Swarm, BYOE, MicroK8s) are you using for SC4S?**

**Is there a pcap available? If so, would you prefer to attach it to this issue or send it to Splunk support?**

**Is the issue related to the environment of the customer or Software related issue?**

**Is it related to Data loss, please explain ?**
Protocol? Hardware specs?

**Last chance index/Fallback index?**

**Is the issue related to local customization?**

**Do we have all the default indexes created?**

**Describe the bug**

**To Reproduce**

mbukowski-splunk:
Hey,
here is the summary:
1. The dev team is running stress tests to verify the issue exists and in what circumstances
2. Checking the SC4S flow and design whether it is switch on/off in the right place.

rjha-splunk:
We have changed the way this feature was implemented , we have tested it with alternate destination enabled.

PR #2536, opened 2024-07-23
fix: parallelize feature is moved to source block

rjha-splunk:
Parallelize feature is moved to source block in place of log_paths


rjha-splunk:
#2534 fixes the issue when using parallelize features, tested after enabling alternate destinations.

mstopa-splunk (approved):

PR #2621, opened 2024-10-17
fix: fix sqlitedict CVE

mstopa-splunk:


package/etc/pylib/parser_vps_cache.py
```diff
@@ -52,7 +86,7 @@ class vpsc_dest(LogDestination):
     def init(self, options):
         self.logger = syslogng.Logger()
         try:
-            self.db = SqliteDict(f"{hostdict}.sqlite", autocommit=True)
+            self.db = SqliteDict(f"{hostdict}.sqlite", autocommit=True, decode=restricted_decode, decode_key=restricted_decode_key)
```
ikheifets-splunk, line 89: It seems too many similar cases, probably have a sense to write a function that will create sqltidict and you can pass *args, **kwargs. Or make a class that will be inherit from SqliteDict and override __init__

mstopa-splunk: done

rjha-splunk (approved):

PR #2637, opened 2024-11-07
Fix/update-helm-charts

wojtekzyla:
Add nodeAffinity and externalTrafficPolicy configuration to values.yaml and change the default value of externalTrafficPolicy to Cluster in order to enable loadbalancing between nodes while using metallb


mstopa-splunk (commented):
It looks like `externalTrafficPolicy: Cluster` obscures the client source IP which we should preserve in the default settings

sbylica-splunk:
Re-opening here:
https://github.com/splunk/splunk-connect-for-syslog/pull/3078

PR #2641, opened 2024-11-18
Non pri cef fix

cwadhwani-splunk:
Updated the regex for non pri cef formated logs such that a '<space>digit' date will be accepted.
Ticket link: https://splunk.atlassian.net/browse/ADDON-76507


package/etc/conf.d/conflib/raw/app-raw-bsd_nopri.conf
```diff
@@ -53,7 +53,7 @@ block parser app-raw-bsd_nopri() {
 };
 application app-raw-bsd_nopri[sc4s-raw-syslog] {
     filter {
-        message('^\w\w\w \d\d \d\d:\d\d:\d\d ');
+        message('^\w\w\w [ \d]\d \d\d:\d\d:\d\d ');
```
rjha-splunk, line 56: Can we make it better regex as well ?

package/lite/etc/conf.d/conflib/raw/app-raw-bsd_nopri.conf
```diff
@@ -53,7 +53,7 @@ block parser app-raw-bsd_nopri() {
 };
 application app-raw-bsd_nopri[sc4s-raw-syslog] {
     filter {
-        message('^\w\w\w \d\d \d\d:\d\d:\d\d ');
+        message('^\w\w\w [ \d]\d \d\d:\d\d:\d\d ');
```
rjha-splunk, line 56: Same comment

rjha-splunk (commented):
Please try to make better regex for dates what i mean by that the dates should be between 1-31 and may be for month as well.

package/etc/conf.d/conflib/raw/app-raw-bsd_nopri.conf
cwadhwani-splunk: Updated the regex with some more constraints.

package/lite/etc/conf.d/conflib/raw/app-raw-bsd_nopri.conf
cwadhwani-splunk: Updated the regex with some more constraints.

PR #2670, opened 2025-01-10
feat: Add a health check endpoint

mstopa-splunk:


rjha-splunk (approved):

package/sbin/healthcheck.py
```diff
@@ -0,0 +1,113 @@
...
+    format=f"%(asctime)s - healthcheck.py - %(levelname)s - %(message)s",
+    datefmt="%Y-%m-%d %H:%M:%S"
+)
+logger = logging.getLogger(__name__)
+
+def check_syslog_ng_health() -> bool:
+    """Check the health of the syslog-ng process."""
+    try:
+        result = subprocess.run(
+            ['syslog-ng-ctl', 'healthcheck', '-t', '1'],
```
ikheifets-splunk, line 33: here timeout 1s, `subprocess.run` timeout 5s

ikheifets-splunk (approved):

package/sbin/healthcheck.py
mstopa-splunk: maybe it's better to keep one longer than the other

syslog-ng-ctl will timeout in 1 second if syslog-ng froze
subprocess will timeout in 5 seconds if anything on the way from Python to syslog-ng-ctl froze

what do you think?

pyproject.toml
```diff
@@ -4,6 +4,7 @@ version = "3.33.1"
 description = ""
 authors = ["rjha-splunk <rjha@splunk.com>"]
 license = "Apache-2.0"
+package-mode = false
```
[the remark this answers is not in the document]
mstopa-splunk: yes

package/sbin/healthcheck.py
ikheifets-splunk: Yeah, I know but our healthcheck is 5s, probably process timeout make 10s. And syslog-ng-ctl timeout 5s

mstopa-splunk: it might be too slow. If an SC4S instance freezes we will be loosing UDP packets for 5 seconds before a LB receives the unhealthy status

ikheifets-splunk: omg, understood :) . I just seen that health check timeout in axosyslog repo for helm chart :) 
Ok, let's stand with that timeout

PR #2678, opened 2025-01-20
Fix poetry export after 2.0

ikheifets-splunk:


package/Dockerfile
```diff
@@ -43,6 +43,8 @@ RUN apk add -U --upgrade --no-cache \
       ca-certificates \
       poetry \
       gdb \
+      py3-poetry-plugin-export \
```
ikheifets-splunk, line 46: In docker image we installing poetry from `apk install`, and [latest version is 1.8.4](https://pkgs.alpinelinux.org/package/edge/community/x86/poetry), by this reason I installing plugin by this way, because when 2.0 will be available for Alpine we will have same problem, that we have on CI

.github/workflows/cd-baremtal.yaml
```diff
@@ -41,7 +41,8 @@ jobs:
         name: Produce BYOE package
         run: |
           pip3 install poetry
-          poetry export   --without-hashes >package/etc/requirements.txt
+          poetry self add poetry-plugin-export
```
ikheifets-splunk, line 44: Starting poetry 2.0 [they made breaking change](https://python-poetry.org/blog/announcing-poetry-2.0.0/#poetry-export-and-poetry-shell-only-available-via-plugins) and `poetry export` will be available only in plugin

mstopa-splunk (approved):

PR #2703, opened 2025-02-12
fix: baremetal release

ikheifets-splunk:


.github/workflows/cd-baremtal.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
           tar rvf /tmp/baremetal.tar -C package/etc .
           tar rvf /tmp/baremetal.tar -C . pyproject.toml
           tar rvf /tmp/baremetal.tar -C . poetry.lock
-          tar rvf /tmp/baremetal.tar -C package/sbin entrypoint.sh
```
ikheifets-splunk, line 49: package/sbin have not only entrypoint.sh file

PR #2690, opened 2025-01-31
fix: resolve sonarqube reported issues

sbylica-splunk:
Fixed sonarqube issues present here:

https://sonarqube.splunkdev.net/dashboard?id=github-mirrors.splunk-connect-for-syslog&branch=sonarcube_issues_fix

PR #2704, opened 2025-02-13
feat: Support multiple destinations for a healthcheck

sbylica-splunk:


.github/workflows/ci-lite.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
         with:
           semantic_version: 18
           extra_plugins: |
-            @semantic-release/exec
+            @semantic-release/exec@v6.0.3
```
sbylica-splunk, line 65: This is fixed in: https://github.com/splunk/splunk-connect-for-syslog/pull/2690 , but we need it for workflow to work properly

package/sbin/healthcheck.py
```diff
@@ -14,11 +15,20 @@ def str_to_bool(value):
...
 
+def get_list_of_destinations():
+    found_destinations = []
+    regex = r"^SC4S_DEST_SPLUNK_HEC_(.*)_URL$"
+
+    for var_key, var_variable in os.environ.items():
+        if re.search(regex, var_key):
+            found_destinations.append(var_variable)
+    return set(found_destinations)
+
```
sbylica-splunk, line 26: This could be imported from: 

https://github.com/splunk/splunk-connect-for-syslog/blob/b8a8c729826d68eebe13b9626f5b7c8eacd9f5f7/package/etc/conf.d/destinations/dest_hec/plugin.py#L28

But a dot in "conf.d" is causing an import issue. Something to think about later I guess

package/sbin/healthcheck.py
```diff
@@ -14,11 +15,20 @@ def str_to_bool(value):
         'yes'
     }
 
+def get_list_of_destinations():
```
ikheifets-splunk, line 18: I just have 2 questions:
* Are we planning to backport same logic from `health check api` to [healthcheck](https://github.com/splunk/splunk-connect-for-syslog/blob/main/package/Dockerfile#L62) 
* Are we planning to add same logic on [entry point](https://github.com/splunk/splunk-connect-for-syslog/blob/main/package/sbin/entrypoint.sh#L186) during startup (where we checking that default destination is healthy, probably we also need to check not only default)


P.S. It's just a question, it doesn't mean that we need to do it or not

.github/workflows/ci-lite.yaml
sbylica-splunk: Merged changes from the develop so it's not needed anymore

package/sbin/healthcheck.py
sbylica-splunk: As for:
1) I think that this is too unrelated. In this dockerfile we are checking for syslog-ng before SC4S is initialised. We "could" also check for HEC health but I'm not sure if this change makes sense.
2) Similar thing really. In this line we are checking mostly for HEC token and default endpoint. I could see a separate check for other HEC endpoints - it could even use the method I just implemented, but it should probably just show warnings.

ikheifets-splunk (approved):
@rjha-splunk @mstopa-splunk can you please take a look on this comment https://github.com/splunk/splunk-connect-for-syslog/pull/2704/files#r1963384996

package/sbin/healthcheck.py
sbylica-splunk: @mstopa-splunk @rjha-splunk - we can talk about this on our sc4s meeting

Issue #2701, opened 2025-02-11
New Parser - Dell PowerStore SCG


zyphermonkey:
**What is the sc4s version?**  
3.34.1  

**Is there a pcap available? If so, would you prefer to attach it to this issue or send it to Splunk support?**  
Log samples provide in support case #3679328

**What the vendor name?**  
Dell

**What's the product name?**  
PowerStore SCG

**If you're requesting support for a new vendor, do you have any preferences regarding the default index and sourcetype for their events?**  

**Do you have syslog documentation or a manual for that device??**  
Product manual available upon request, but it does not appear to contain any information in regards to the syslog format/template.  

**Feature Request description:**

**Do you want to have it for local usage or prepare a github PR?**  
github PR preferred, but local usage is also an acceptable solution.

zyphermonkey:
> Hello, the parser is released in v3.37.0

Thx. 
I believe these should default to index `infraops` instead of `netops`. These are events from storage appliances and not networking devices.

PR #2770, opened 2025-06-17
fix: update dell powerstore index

ajasnosz:
Updating the default index as the events are coming from storage appliances. Based on comment from: #2701


package/enterprise/etc/conf.d/conflib/syslog/app-syslog-dell_powerstore.conf
```diff
@@ -2,7 +2,7 @@ block parser app-syslog-dell_powerstore() {
  channel {
         rewrite {
             r_set_splunk_dest_default(
-                index('netops')
+                index('infraops')
```
mstopa-splunk: Please check if we can update enterprise with this PR

ajasnosz: As we stop the updates for enterprise, I have removed the changes.

PR #2801, opened 2025-10-13
fix: new release v3.38.1

pdudhaiya-crest:
PR to merge develop to main an trigger release.

Changes in this release:

- fix: update requests and urllib3 to fix vulnerabilities
- Fixed sonarqube issues
- Added docs for EKS deployment
- Add end of support notice for BYOE deployments

### DIFF
--- a/docs/gettingstarted/byoe-rhel8.md
+++ b/docs/gettingstarted/byoe-rhel8.md
@@ -1,5 +1,11 @@
 # Configure SC4S in a non-containerized SC4S deployment
 
+!!! warning ""
+    Support for all non-containerized deployments using the "bring your own environment" method, as detailed in this documentation, will end in **April 2026**.
+    After that, **no further updates, security patches, or technical support will be provided for this deployment method**. Continued use may expose your environment to security risks and compatibility issues.
+    
+    We recommend planning your migration to one of the other supported platform or deployment method before April 2026 to ensure continued support and security. See [Select a Container Runtime and SC4S Configuration](https://splunk.github.io/splunk-connect-for-syslog/main/gettingstarted/getting-started-runtime-configuration/#step-3-select-a-container-runtime-and-sc4s-configuration) for more details.
+
 Configuring SC4S in a non-containerized SC4S deployment requires a custom configuration. Note that since Splunk does not control your unique environment, we cannot help with setting up environments, debugging networking, etc. Consider this configuration only if:
 
 * Your specific requirements preclude the use of containers or demand that you use SC4S in a custom environment.
--- a/docs/gettingstarted/getting-started-runtime-configuration.md
+++ b/docs/gettingstarted/getting-started-runtime-configuration.md
@@ -69,7 +69,7 @@ Check your Podman or Docker documentation to see which operating systems are sup
 | [Docker CE + systemd](docker-systemd-general.md)                  |                                                                                     |
 | [Docker Desktop + Compose](docker-compose-MacOS.md)               | MacOS                                                                               |
 | [Docker Compose](docker-compose.md)                               |                                                                                     |
-| [Bring your own Environment](byoe-rhel8.md)                       | RHEL or CentOS 8.1 & 8.2 (best option)                                              |
+| (Deprecated) [Bring your own Environment](byoe-rhel8.md)          | RHEL or CentOS 8.1 & 8.2 (best option)                                              |
 | [Offline Container Installation](docker-podman-offline.md)        |                                                                                     |
 | [Ansible+Docker Swarm](ansible-docker-swarm.md)                   |                                                                                     |
 | [Ansible+Podman](ansible-docker-swarm.md)                         |                                                                                     |
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -24,6 +24,7 @@ markdown_extensions:
           format: !!python/name:pymdownx.superfences.fence_code_format
   - pymdownx.blocks.tab:
       alternate_style: true
+  - admonition
 
 
 theme:


pdudhaiya-crest:
@ajasnosz , There are 2 stages failing - 

for security-fossa-scan, there are only licensing issues, so I will raise a ticket on tonkean with legal team to mark it false positive.

for "Scan docker images on CVEs", it does not seem to block release and I see that we had not fixed it for some releases : https://github.com/splunk/splunk-connect-for-syslog/actions/runs/16723983695

charts/splunk-connect-for-syslog/Chart.yaml
```diff
@@ -3,5 +3,5 @@ apiVersion: v2
 name: splunk-connect-for-syslog
 description: Deploy Splunk Connect for Syslog
 type: application
-version: 3.37.0
-appVersion: "3.37.0"
+version: 3.38.0
+appVersion: "3.38.0"
```
sbylica-splunk, line 7: This should be 3.38.1 right?

ajasnosz: The pipeline is changing the versioning after merge

.github/workflows/ci-lite.yaml
```diff
@@ -383,7 +383,7 @@ jobs:
       - name: Return status 'healthy'
         run: python3 tests/test_healthcheck_healthy.py --host sc4s --port 8090
       - name: Return status 'queue size exceeded limit'
-        run: python3 tests/test_healthcheck_queue_size_limit.py --limit 10000 --host sc4s --port 8090
+        run: python3 tests/test_healthcheck_queue_size_limit.py --limit 20000 --host sc4s --port 8090
```
Kawron, line 386: Why we decided to increase it to 2000? Is there any reason for choosing this number?

Also the naming of this argument is a little bit missleading, if you look in the code and logs you will see:
`print(f"Sending {args.limit} messages to {args.host}:{args.udp_port}...")`
so the limit is not the queue limit but the amount of messages send. The default limit is 10000 as seen in file package/sbin/healthcheck.py:
`MAX_QUEUE_SIZE = int(os.getenv('HEALTHCHECK_MAX_QUEUE_SIZE', '10000'))`

This tests exepcts that the queue limit will be exceeded. If this tests fails that means that the number of queued messages is less then 10000.

Now I'm not entirerly sure whether increasing this arg is a bad thing. But I just want to make sure that we are making an informed decision, as this looks a little bit sus at the first look.

Kawron: Also, can we be sure that the fact that the queue has less then 10000 messages is due to fact that the queue is being processed to quickly or that we are loosing some messages?

Kawron: And can we identify the core reason for this issue? It doesn't fail in every PR, is this non-determistic or maybe can we pinpoint the exact cause of this issue?

pdudhaiya-crest: @Kawron 

`What is the reason for the failing healthcheck tests and why we decided to increase it to 2000?`
--> healthcheck test compares the current queue size by running "syslog-ng-ctl stats" (https://github.com/splunk/splunk-connect-for-syslog/blob/develop/package/sbin/healthcheck.py#L78)  and the configured limit for queue size in SC4S via the variable HEALTHCHECK_MAX_QUEUE_SIZE (https://github.com/splunk/splunk-connect-for-syslog/blob/develop/package/sbin/healthcheck.py#L34). 

Peviously in the tests, the limit was 10000 and we were testing the queue limit also by sending 10000 test messages. In this case, the tests were failing intermittently (flaky). I had tried 10001 and the tests continued to fail intermittently. It might be the case that SC4S is consuming messages in the queue continuously and that's why there are not more that 10000 messages in the queue at a particular time and that's why the tests were failing. 

`Is there any reason for choosing this number?`
--> Since it seems that these tests are meant to check the Queue limit and report an unhealthy status on the server, it is good to give a sufficiently large number of messages in the queue so that the case is tested everytime without fail.


`Also the naming of this argument is a little bit missleading`
--> I agree, the naming is not intuitive. We can change the names of variables and script arguments to best reflect the intention.

Kawron: What worries me is that this wasn't a case before, I'm looking a at the previous builts trying to pinpoint the change that made those tests fail almost constantly, but for know I cannot find it.

`It might be the case that SC4S is consuming messages in the queue continuously and that's why there are not more that 10000 messages in the queue at a particular time and that's why the tests were failing.`
--> If this is the case, that would be great it means that we've increased our ingestion rates. Neverthless, the tests are supposed to alert us about unexpected changes, this for sure is unexpected and for now unexplainable. 

I think, that we should modify this tests, to log the exact amount of messages in the queue so we can track this behaviour in the feature. Though this can be done in a seperate PR.

Kawron (approved):

Kawron (approved):

.github/workflows/ci-lite.yaml
Kawron: When it comes to merging this PR, this was already an issue in the previous version 3.38.0, so we won't break anything new. But If we are gonna merge it, I would personally leave the tests failing so we won't obfuscate the tests, until we know what is the core issue

pdudhaiya-crest: The problem with leaving these tests failing is that -- we won't be able to trigger a release since this is a blocking test. Do we want to move ahead with this OR mark these tests as non blocking and allow them to fail for now.

@ajasnosz @sbylica-splunk , let me know what you guys suggest.

pdudhaiya-crest: I think we should move ahead for now and if this is a concerning test, we can continue investigating this behaviour after this release.

ajasnosz: I remember the test was flaky for a long time already, so I think we can take a note to improving how it works and consider if this is ok to increase the numbers or if we could redesign and test this in some other way.

pdudhaiya-crest: I have created this task for addressing this : https://splunk.atlassian.net/browse/ADDON-84183

WIll move ahead with the release for now.

Issue #2796, opened 2025-09-17
Bug in barracuda_waf parser


ehlo550:
**What is the sc4s version ?**
3.37.0

**Describe the bug**
As mentioned in the slack community.
When trying to onboard syslogs from [Cisco CIMC ](https://www.cisco.com/c/en/us/td/docs/unified_computing/ucs/c/sw/monitoring/guide/b_Monitoring_Cisco_IMC_Using_Syslog/b_Monitoring_Cisco_IMC_Using_Syslog_chapter_010.html#id_83377) the logs are identified as barracuda_waf due to the program being AUDIT.

Sample Log from Cisco CIMC:
```
<37>Sep 16 10:51:14 hostname001 AUDIT[2386]: Login failed (ip:192.168.12.230, service:webgui) due to invalid password
```

sbylica-splunk:
Can we get the pcap file / user configuration?

On a quick glance it doesn't look like a proper CIMC log, you can check the proper structure here:
https://www.cisco.com/c/en/us/td/docs/unified_computing/ucs/c/sw/monitoring/guide/b_Monitoring_Cisco_IMC_Using_Syslog/b_Monitoring_Cisco_IMC_Using_Syslog_chapter_01.pdf

sbylica-splunk:
Hi @ehlo550 sorry for a late reply, I was pulled into another case. 
Yeah if you can produce a fault which would create a valid %CIMC log it would help me. 
Also do you have access to local logs by chance? We could try to compare them with the ones send to sc4s.

Regards, Szymon

sbylica-splunk:
Hi @ehlo550 I can join a meeting this Thursday or some other day next week, 9:00AM - 5:00PM Central European Time.
Do you have any preferred timeslot?

ehlo550:
Hi, 
I would have time for example between: 
07 Oct: 1300 - 1600
08 Oct: 0900 - 1600
Regards
Stefan

sbylica-splunk:
@ehlo550 

Lets say 08 Oct at 13:00 CET, is this ok for you?

Regards, Szymon

sbylica-splunk:
New version of SC4S with this fix just got released, closing this issue.
https://github.com/splunk/splunk-connect-for-syslog/releases/tag/v3.39.0

PR #2812, opened 2025-10-27
fix: Barracuda parser adjustment

sbylica-splunk:
Fixed issue raised in: https://github.com/splunk/splunk-connect-for-syslog/issues/2796


sbylica-splunk:
> I'm not sure what happened in the commit tree that there are some other commits visible, but could you try to rebase or squash and rebase to update the commit name with the semantic-release format? And remove the commits with confusing names like `.`

I can squash and merge at the same time so I'll do it then.

PR #2823, opened 2025-11-05
test: (ADDON-81912) add docker and podman e2e test strategy

pdudhaiya-crest:
Jira: https://splunk.atlassian.net/browse/ADDON-81912 

* Added new e2e test strategy for docker and podman
* Added artifacts for each tests and matrix

**NOTE**: 1 tests case is failing for "test-container" and "test-name-cache". I am working on fixing thise


.github/workflows/ci-main.yaml
```diff
@@ -437,8 +456,7 @@ jobs:
       - meta
       - build_action
       - test-container
-      - test-ipv4-name-cache
-      - test-ipv6-name-cache
+      - test-name-cache
```
sbylica-splunk, line 428: Looks like we have some tests failing now:
`[gw0] [ 20%] FAILED tests/test_name_cache.py::test_name_cache `

pdudhaiya-crest: Yes, as mentioned in the PR note, there is 1 test failing for each stages. I am checking on it, but this is to review the rest of the changes  of the Github Actions.

.github/workflows/ci-enterprise.yaml
```diff
@@ -202,41 +205,68 @@ jobs:
...
         uses: actions/checkout@v4
         with:
           submodules: false
           persist-credentials: false
+      - name: Install Ansible and other dependencies as python package
+        run: |
+          pip install ansible~=6.1.0 --no-cache-dir \
+          && pip install pywinrm>=0.4.2 --no-cache-dir \
+          && pip install ansible-lint>=6.0.0 --no-cache-dir \
+          && pip install docker
```
Kawron: if we are using --no-cache-dir, I would use it for docker installment as well

ansible/tasks/podman/deploy_app.yml
```diff
@@ -12,15 +12,15 @@
 
 - name: Copying unit file on the server
   copy:
-    src: "/opt/ansible/resources/podman_sc4s.service"
+    src: "../../../ansible/resources/podman_sc4s.service"
```
Kawron, line 15: why cannot we use absolute path? Using relative path is more error prone

.github/workflows/ci-main.yaml
```diff
@@ -178,14 +178,14 @@ jobs:
           scanners: "vuln"
           
   test-container:
+    strategy:
+      matrix: 
+        deployment_type: ["docker", "podman"]
     runs-on: ubuntu-latest
+    continue-on-error: true
```
Kawron, line 185: is there any reason for adding this?

ajasnosz (commented):
I don't have other comments that weren't already mentioned here, if they will be addressed and the test would pass I think it's ok.

For the future PRs we also use tags from semantic release like fix/ test/ instead of addon number in the branch name 😃

nikunjj-crest (approved):

.github/workflows/ci-main.yaml
pdudhaiya-crest: without this, if any one of the runs in the matrix fails, all others are skipped. So if tests running on podman fails, the tests running on docker will terminate/skip in middle of its execution.

ansible/tasks/podman/deploy_app.yml
pdudhaiya-crest: @Kawron  The PWD when running the ansible playbooks is this in our case on CI -- 

`"current_directory_output.stdout": "/home/runner/work/splunk-connect-for-syslog/splunk-connect-for-syslog/ansible/playbooks"`

Also, this can be anywhere where the repo is copied but the relative path should work.
In the current state, the ansible playbook is not working and giving this error : 

```
TASK [Copying unit file on the server] ******************************************************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: If you are using a module and expect the file to exist on the remote, see the remote_src option
fatal: [node_1]: FAILED! => {"changed": false, "msg": "Could not find or access '/opt/ansible/resources/docker_sc4s.service' on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}

```

.github/workflows/ci-enterprise.yaml
pdudhaiya-crest: Made the changes. Thanks!

PR #2834, opened 2025-12-17
feat: configurator tool script

sbylica-splunk:


docs/architecture/index.md
```diff
@@ -54,3 +54,25 @@ The following resources can help you determine the best protocol for your setup:
 
 1. [Run performance tests for TCP](performance-tests.md#check-your-tcp-performance)
 2. [Run performance tests for UDP](performance-tests.md#check-your-udp-performance)
+
+### Configuration helper for SC4S
+
+SC4S includes an interactive configuration tool that simplifies the setup and optimization of your environment file. The tool walks you through key configuration decisions and generates a customized `env_file` based on your specific requirements.
```
ajasnosz, line 60: I'm thinking if it would be good option to record how running the script looks like, and also share the video of it here?

configuration-tool.sh
```diff
@@ -0,0 +1,344 @@
...
+        else
+            read -p "$question [y/N]: " response
+            response=${response:-n}
+        fi
+        
+        case "$response" in
+            [Yy]* ) echo "yes"; break;;
+            [Nn]* ) echo "no"; break;;
+            * ) echo "Please answer yes or no.";;
+        esac
```
ajasnosz, line 71: I guess we could expand that to "Y y yes Yes" if somebody would type the whole word, similar for "no".

configuration-tool.sh
```diff
@@ -0,0 +1,344 @@
...
+PROTOCOL="both"
+
+SC4S_SOURCE_LISTEN_UDP_SOCKETS=2
+SC4S_SOURCE_UDP_FETCH_LIMIT=1000
+SC4S_ENABLE_EBPF="no"
+SC4S_EBPF_NO_SOCKETS=4
+
+PARALLELIZE="no"
+SC4S_PARALLELIZE_NO_PARTITION=4
+SC4S_SOURCE_TCP_IW_USE="no"
```
Kawron, line 31: https://github.com/search?q=repo%3Asplunk%2Fsplunk-connect-for-syslog%20IW_USE&type=code

According to this search, such options does not exist

configuration-tool.sh
```diff
@@ -0,0 +1,344 @@
...
+SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_MEMBUFLENGTH=15000
+SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_DISKBUFSIZE=53687091200
+
+echo ""
+echo "${BLUE}================================================${NC}"
+echo "${BLUE}    SC4S Configuration Tool${NC}"
+echo "${BLUE}================================================${NC}"
+echo ""
+echo "This tool will help you generate an optimized SC4S configuration"
+echo "based on your environment requirements."
```
ajasnosz: Not sure if this is the right description for now, as we don't propose any optimization based on the env, but rather for now the customer is configuring them. I guess this would be provided in next tasks?

ajasnosz (changes requested):

docs/architecture/index.md
sbylica-splunk: I'm not really sure about that, we don't use videos in this documentation and this is a fairly simple script

configuration-tool.sh
sbylica-splunk: Yeah, makes sense

configuration-tool.sh
sbylica-splunk: Ok, checked it and seems like this is the case. I'll keep this variable for the purposes of the script but it will not be added to the generated file

ajasnosz (approved):

PR #2861, opened 2025-12-23
Chore/update setuptools again

sbylica-splunk:


sbylica-splunk:
Looks like updating setuptools to v80 creates a problem with fossa, reverting for now

PR #2864, opened 2026-01-12
feat: Sc4s configurator

sbylica-splunk:
Configurator script that can be used to help with creating of env_file

### DIFF
--- a/docs/architecture/index.md
+++ b/docs/architecture/index.md
@@ -75,4 +75,6 @@ sh ./configuration-tool.sh
 
 The tool will prompt you for your environment details and create an optimized `env_file` ready for deployment. This approach ensures consistent configuration and helps you apply best practices without manually editing environment variables.
 
+You can also use this tool to create a suggested configuration, based on your hardware and expected number of events per second (EPS).  To do that select the "Hardware-based configuration" mode when prompted.
+
 You can also read about our [scaling solutions](performance-tests.md) and [configuration variables](../configuration.md).


configuration-tool.sh
```diff
@@ -72,15 +72,149 @@ ask_yes_no() {
...
+if [[ "$mode_choice" == "2" ]]; then
+    # Hardware-based mode
+    echo ""
+    echo "${GREEN}=== Hardware-Based Configuration ===${NC}"
+    echo ""
+    echo "Select type of instance most similar to your hardware:"
+    echo "1) m5.4xlarge (16 vCPUs, 64 GB RAM)"
+    echo "2) m5.2xlarge (8 vCPUs, 32 GB RAM)"
+    echo "3) m5.xlarge (4 vCPUs, 16 GB RAM)"
+    echo ""
```
ajasnosz, line 164: I think we shouldn't suggest the naming convention from aws, but just specify by cpus and ram. 
I'm also wondering if maybe it would be better that we ask the customer to type their cpu and ram, and based on that we would select which option would be the best?

configuration-tool.sh
```diff
@@ -72,15 +72,149 @@ ask_yes_no() {
...
+    read -p "Select hardware [2]: " hw_choice
+    hw_choice=${hw_choice:-2}
+    
+    case "$hw_choice" in
+        1) HARDWARE="m5.4xlarge";;
+        2) HARDWARE="m5.2xlarge";;
+        3) HARDWARE="m5.xlarge";;
+        *) HARDWARE="m5.2xlarge";;
```
ajasnosz: similarly I wouldn't recommend to use internally in the script the aws naming convention for ec2, but also something compatible with other envs like cpu ram

Kawron (approved):

configuration-tool.sh
sbylica-splunk: I think I can invert this and put (like m5.4xlarge EC2) at the end, it should be cleaner

ajasnosz (approved):

PR #2866, opened 2026-01-12
doc: Update the test results with new data and update the finetuning …

Kawron:
Update the test results with new data and update the finetuning information based on newly collected test results

### DIFF
--- a/docs/architecture/fine-tuning.md
+++ b/docs/architecture/fine-tuning.md
@@ -66,7 +66,31 @@ SC4S_SOURCE_RFC5425_SO_RCVBUFF=536870912   # RFC 5425 (syslog over TLS)
 
 ### Impact on performance
 
-According to tests performed in our lab environment, tuning the receive buffer showed the best results for TCP connections, where it doubled message throughput in some cases. For UDP connections, the impact was less noticeable, especially for larger throughputs. However, two key things were observed: first, the best results were achieved using a 64 MB receive buffer; second, according to syslog documentation and blog posts, setting the receive buffer to a higher value decreases the probability of small UDP losses for relatively low message throughput. In our environment, we observed losses around 0.2% at 9,000 msg/sec throughput. These losses disappeared after increasing the receive buffer to 64 MB. Because of this, it is recommended to start with a relatively small value for UDP (for example, the default 16 MB or 64 MB) and increase if necessary after trying other options like eBPF.
+Test performed in the lab environment show that tuning receive buffer can improve performance, depending on the connection type.
+
+**TCP connections**
+
+* Receive buffer tuning produced the best results for TCP connections.
+* In some cases, message throughput doubled.
+
+**UDP connections**
+The performance impact was smaller, especially at higher throughputs.
+
+Two important observations were made:
+
+* The best results are achieved with a 64 MB receive buffer.
+* According to syslog documentation and blog posts, higher receive buffer values reduce the probability of small UDP packet losses at relatively low message throughput.
+
+Observed behaviour:
+
+* In our environment, we observed message losses of approximately 0.2% at a throughput of 9,000 msg/sec.
+* These losses disappeared after increasing the receive buffer to 64 MB.
+
+For UDP connections you should:
+
+* Start with a relatively small receive buffer value (for example, the default 16 MB or 64 MB).
+* Increase the value only if needed.
+* First try other tuning options, such as eBPF, before increasing receive buffer.
 
 ## Tune static input window size
 
@@ -101,13 +125,13 @@ For **UDP**, if the output becomes slow and this window fills up, syslog‑ng wi
 
 A single UDP message can be up to approximately 1 KB. With a window size of 1,000,000 messages, this may require up to **1 GB** of additional memory for buffering.
 
-### Impact on performance
+**Impact on performance**
 
-The input window option can greatly improve performance for burst loads. Increasing the input window size won't improve the baseline throughput of SC4S, but it will allow ingesting an increased rate of messages for a limited time until the input window fills up. Keep in mind that this will also add latency to the time of events arriving in Splunk.
+If you increase the window size, you will not improve the baseline throughput of SC4S, but it will allow ingesting an increased rate of messages for a limited time until the input window fills up. Keep in mind that this will also add latency to the time of events arriving in Splunk.
 
 For example, in our UDP test scenario on an EC2 machine with 16 threads and eBPF enabled, with 150,000 messages per second arriving over the span of 30 seconds, we managed to reduce the 50% loss rate to 0% by setting input window to 1000000 (1 GB). Similarly, for the test scenario with 350,000 messages per second, without input window we achieved a loss rate of around 80%, which corresponds to only 75,000 messages per second. With input window enabled, we managed to lower this to 61.49%, which is almost 150,000 messages per second. However, once the buffer fills up, there will be no improvement in performance.
 
-The default value for TCP input window size is already large enough and in most cases doesn't require further tuning.
+The default value for TCP input window size is already large enough and in most cases does not require further tuning.
 
 ### Fetch limit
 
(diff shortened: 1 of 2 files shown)


docs/architecture/fine-tuning.md
```diff
@@ -64,6 +64,10 @@ SC4S_SOURCE_RFC5425_SO_RCVBUFF=536870912   # RFC 5425 (syslog over TLS)
...
 - **Hardware constraints:** Network driver limitations should be considered when tuning these values. Consult your NIC documentation for maximum supported buffer sizes.
 
+### Impact on performance
+
+According to tests performed in our lab environment, tuning the receive buffer showed the best results for TCP connections, where it doubled message throughput in some cases. For UDP connections, the impact was less noticeable, especially for larger throughputs. However, two key things were observed: first, the best results were achieved using a 64 MB receive buffer; second, according to syslog documentation and blog posts, setting the receive buffer to a higher value decreases the probability of small UDP losses for relatively low message throughput. In our environment, we observed losses around 0.2% at 9,000 msg/sec throughput. These losses disappeared after increasing the receive buffer to 64 MB. Because of this, it is recommended to start with a relatively small value for UDP (for example, the default 16 MB or 64 MB) and increase if necessary after trying other options like eBPF.
```
ajasnosz: could you ut the first and second parts on the bullet points? I think that would improve readability.

docs/architecture/fine-tuning.md
```diff
@@ -95,6 +101,14 @@ For **UDP**, if the output becomes slow and this window fills up, syslog‑ng wi
 
 A single UDP message can be up to approximately 1 KB. With a window size of 1,000,000 messages, this may require up to **1 GB** of additional memory for buffering.
 
+### Impact on performance
+
+The input window option can greatly improve performance for burst loads. Increasing the input window size won't improve the baseline throughput of SC4S, but it will allow ingesting an increased rate of messages for a limited time until the input window fills up. Keep in mind that this will also add latency to the time of events arriving in Splunk.
```
ajasnosz: will not

docs/architecture/fine-tuning.md
```diff
@@ -95,6 +101,14 @@ For **UDP**, if the output becomes slow and this window fills up, syslog‑ng wi
...
+
+For example, in our UDP test scenario on an EC2 machine with 16 threads and eBPF enabled, with 150,000 messages per second arriving over the span of 30 seconds, we managed to reduce the 50% loss rate to 0% by setting input window to 1000000 (1 GB). Similarly, for the test scenario with 350,000 messages per second, without input window we achieved a loss rate of around 80%, which corresponds to only 75,000 messages per second. With input window enabled, we managed to lower this to 61.49%, which is almost 150,000 messages per second. However, once the buffer fills up, there will be no improvement in performance.
+
+The default value for TCP input window size is already large enough and in most cases doesn't require further tuning.
```
ajasnosz: does not

docs/architecture/performance-tests.md
```diff
@@ -48,96 +48,85 @@ Usage:
...
+* Splunk Enterprise - m5.4xlarge single indexer
 
-### Command
+The *"Finetuned"* configuration for 1 active connection used the following settings:
 ```bash
-/opt/syslog-ng/bin/loggen -i --rate=100000 --interval=1800 -P -F --sdata="[test name=\"stress17\"]" -s 800 --active-connections=10 <local_hostmane> <sc4s_external_tcp514_port>
+SC4S_ENABLE_PARALLELIZE=yes
+SC4S_PARALLELIZE_NO_PARTITION=16 # number of cores
 ```
+The **SC4S_ENABLE_PARALLELIZE** setting doesn't increase performance when multiple TCP connections are used. For more information, refer to [Finetune SC4S for TCP](fine-tuning.md#finetune-for-tcp-traffic).
```
ajasnosz: does not

docs/architecture/performance-tests.md
```diff
@@ -48,96 +48,85 @@ Usage:
...
+sudo netstat -ausn
 ```
+The number of errors should match the number of missing messages in Splunk.
 
-Over a span of 60 seconds, loggen will attempt to generate 20,000 logs per second, each 800 bytes in size, which will be sent via UDP.
+## Watch out for queues
+Comparing loggen results can be sufficient for A/B testing, but is not adequate for estimating the syslog ingestion throughput of the entire system.
 
-After running the command, count the number of events that reached Splunk. Since UDP is prone to data loss, messages can be lost anywhere along the path.
+In the following example, loggen was able to send 4.3 mln messages in one minute; however, Splunk indexers required an additional two minutes to process these messages. During that time, SC4S processed the messages and stored them in a queue while waiting for the HEC endpoint to accept new batches.
```
ajasnosz: million

docs/architecture/fine-tuning.md
jplatosz-splunk: You can rewrite it as follows:

Test performed in the lab environment show that tuning receive buffer can improve performance, depending on the connection type.

**TCP connections**
- Receiver buffer tuning produced the best results for TCP connections.
- In some cases, message throughput doubled.

**UDP connections**
The performance impact was smaller, especially at higher throughputs.

Two important observations were made:
- The best results are achieved with a 64 MB receive buffer
- According to syslog documentation and blog posts, higher receive buffer values reduce the probability of small UDP packet losses at low message throughput.

**Observed behavior**
- In our environment, we observed message losses of approximately 0.2% at a throughput of 9,000 msg/sec.
- These losses disappeared after increasing the receive buffer to 64 MB.

For UDP connections you should:
- Start with a relatively small receive buffer value (for example, the default 16 MB or 64 MB).
- Increase the value only if needed.
- First try other tuning options, such as eBPF, before increasing receive buffer.

docs/architecture/fine-tuning.md
jplatosz-splunk: If you increase the window size, you will not improve the baseline (...)

docs/architecture/fine-tuning.md
```diff
@@ -159,6 +158,10 @@ SC4S_SOURCE_LISTEN_UDP_SOCKETS=32
 
 Set this value based on the number of CPU cores available. Start with a value equal to your 4 x core count and adjust based on performance testing. Restart SC4S for the changes to take effect.
 
+#### Impact on the performance
```
jplatosz-splunk: We avoid this heading density to maintain readability. You can use bold instead **Impact on the performance**

docs/architecture/fine-tuning.md
```diff
@@ -159,6 +158,10 @@ SC4S_SOURCE_LISTEN_UDP_SOCKETS=32
...
+
+Increasing the number of UDP sockets yields the best results in scenarios where data comes from different sources, as it allows utilizing more CPU cores. If the data comes only from a single source (same IP and port), this option will have a marginal impact on performance. In that case, consider enabling eBPF.
```
jplatosz-splunk: Increasing the number of UDP sockets provides the best results when data comes from multiple sources, as it allows better utilization of CPU cores.

If the data comes from a single source (same IP and port), this setting has only a minimal impact on performance. In this case, consider enabling eBPF instead.

docs/architecture/fine-tuning.md
```diff
@@ -169,26 +172,27 @@ Find more in the [About eBPF](../../configuration/#about-ebpf) section.
 ```bash
 SC4S_SOURCE_LISTEN_UDP_SOCKETS=32
 SC4S_ENABLE_EBPF=yes
-SC4S_EBPF_NO_SOCKETS=32
+SC4S_EBPF_NO_SOCKETS=32 # to achieve best results set to number of threads x 4
 ```
 4. Restart SC4S for the changes to take effect.
 
-## Finetune for TCP traffic
+#### Impact on performance
```
jplatosz-splunk: see my earlier comment about headings :)

docs/architecture/fine-tuning.md
```diff
@@ -169,26 +172,27 @@ Find more in the [About eBPF](../../configuration/#about-ebpf) section.
...
 
-### Tested configuration:
-- **Loggen** - c5.2xlarge
-- **SC4S** (3.29.0) + podman - c5.4xlarge
-- **Splunk Cloud** 9.2.2403.105 - 30IDX
+Depending on which machine you are running SC4S on, this option can greatly increase performance. The best results will be achieved on multithreaded machines. This is demonstrated in the performance test results for different EC2 instances (m5.4xlarge, m5.2xlarge, and m5.xlarge):
```
jplatosz-splunk: You can achieve the best results on multithreaded machines.

jplatosz-splunk (changes requested):
left a few comments and suggestions

docs/architecture/fine-tuning.md
Kawron: Followed Justyna sugestions https://github.com/splunk/splunk-connect-for-syslog/pull/2866#discussion_r2685777414

PR #2879, opened 2026-01-20
fix: avoid double --no-caps cli option for syslog-ng

Klaas-:
Currently --no-caps is set twice, once in package/sbin/entrypoint.sh once in SC4S_CONTAINER_OPTS inside package/Dockerfile*

I've removed it in entrypoint so it gets set via SC4S_CONTAINER_OPTS and is overwriteable if anyone choses to do so.

### DIFF
--- a/package/sbin/entrypoint.sh
+++ b/package/sbin/entrypoint.sh
@@ -249,7 +249,7 @@ done
 export SC4S_VERSION=$(cat $SC4S_ETC/VERSION)
 echo sc4s version=$(cat $SC4S_ETC/VERSION)
 echo sc4s version=$(cat $SC4S_ETC/VERSION) >>$SC4S_VAR/log/syslog-ng.out
-"${SC4S_SBIN}"/syslog-ng --no-caps $SC4S_CONTAINER_OPTS -s >>$SC4S_VAR/log/syslog-ng.out 2>$SC4S_VAR/log/syslog-ng.err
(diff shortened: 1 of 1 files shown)


package/sbin/entrypoint.sh
```diff
@@ -279,12 +279,7 @@ fi
 while :
 do
   echo starting syslog-ng
-  if [ "${SC4S_DEBUG_LOGS}" == "yes" ]; then
-    echo debug mode enabled
-    "${SC4S_SBIN}"/syslog-ng --no-caps -d -v -e "${SC4S_CONTAINER_OPTS}" -F "${@}" &
-  else
-    "${SC4S_SBIN}"/syslog-ng --no-caps "${SC4S_CONTAINER_OPTS}" -F "${@}" &
-  fi
+  "${SC4S_SBIN}"/syslog-ng "${SC4S_CONTAINER_OPTS}" -F "${@}" &
```
Kawron: It seems like merge conflict, but anyway there is no reason to remove the SC4S_DEBUG_LOGS if statment. Please make sure to pull latest changes

Klaas-: Yeah this is showing like this because the part I changed was also changed in develop branch after I made the PR. I'll rebase it and remove the --no-caps from both if paths

Klaas-: ah no, this is because of main/develop -- I'll switch to develop as base for my PR

PR #2880, opened 2026-01-20
fix: add formating for LogMessage and add debug log mode

Kawron:
Adding flag for enabling the debug mode for syslog-ng logger.

Fixing typerrors, by adding log utility functions.

After consideration desided not to seperate the sc4s logs and syslog logs due to the cost of maintaining this solution. Some of our parser extend syslog python library which could results in issues in the future. For now it will be recommended to filter the logs with grep or other tools.

### DIFF
--- a/docs/troubleshooting/troubleshoot_SC4S_server.md
+++ b/docs/troubleshooting/troubleshoot_SC4S_server.md
@@ -176,7 +176,8 @@ SC4S_DEBUG_LOGS=yes
 
 This runs syslog-ng with debug (`-d`), verbose (`-v`), and stderr (`-e`) flags. When enabled internal syslog-ng debug messages appear in `docker/podman logs`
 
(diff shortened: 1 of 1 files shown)


docs/troubleshooting/troubleshoot_SC4S_server.md
```diff
@@ -166,6 +166,18 @@ In this example the error can be seen in the snippet `statefulset.kubernetes.io/
...
+
+To enable verbose debug logging that outputs to stderr (visible in `docker/podman logs`), set the following property in `env_file`:
+
+```
+SC4S_DEBUG_LOGS=yes
+```
+
+This runs syslog-ng with debug (`-d`), verbose (`-v`), and stderr (`-e`) flags. When enabled internal syslog-ng debug messages appear in `docker/podman logs`
+
+**Note:** Debug logging increases output verbosity significantly. Use only for troubleshooting, not in production.
```
ajasnosz: Instead of **note**  let's use
!!! note

PR #2894, opened 2026-01-27
Feat: add syslog pcap sender

ajasnosz:
`syslog_pcap_sender` extracts the syslog payloads and sends them over a fresh TCP or UDP connection to syslog server.

### DIFF
--- a/syslog_pcap_sender.py
+++ b/syslog_pcap_sender.py
@@ -343,54 +343,6 @@ class SyslogSender:
             sys.exit(1)
 
         return self.payloads
-    # def _extract_udp_payloads(self):
-    #     """
-    #     Extract payloads from UDP packets (one message per packet).
-    #     """
-    #     if self.no_dedup:
-    #         print("  (Deduplication disabled - all payloads will be sent)")
-    #
-    #     seen_payloads = set()
-    #     packets_with_payload = 0
-    #     duplicate_count = 0
-    #
-    #     for pkt in self.packets:
-    #         if UDP in pkt and Raw in pkt:
-    #             payload = bytes(pkt[Raw].load)
-    #             packets_with_payload += 1
-    #
-    #             if payload and len(payload) > 0:
-    #                 if self.no_dedup:
-    #                     self.payloads.append(payload)
-    #                 else:
-    #                     payload_hash = hash(payload)
-    #                     if payload_hash not in seen_payloads:
-    #                         seen_payloads.add(payload_hash)
-    #                         self.payloads.append(payload)
-    #                     else:
-    #                         duplicate_count += 1
-    #
-    #     if self.no_dedup:
-    #         print(f"Extracted {len(self.payloads)} syslog messages (including duplicates)")
-    #     else:
-    #         print(f"Extracted {len(self.payloads)} unique syslog messages")
-    #
-    #     # Show filtering details
-    #     udp_packets = sum(1 for p in self.packets if UDP in p)
-    #     packets_without_payload = udp_packets - packets_with_payload
-    #
-    #     if packets_without_payload > 0 or (duplicate_count > 0 and not self.no_dedup):
-    #         print(f"  Filtering: {len(self.packets)} total packets → {len(self.payloads)} messages")
-    #         if packets_without_payload > 0:
-    #             print(f"    - {packets_without_payload} UDP packets without payload")
-    #         if duplicate_count > 0 and not self.no_dedup:
-    #             print(f"    - {duplicate_count} duplicate payloads (deduplicated)")
-    #
-    #     if len(self.payloads) == 0:
-    #         print("No syslog payloads found in packets", file=sys.stderr)
-    #         sys.exit(1)
-    #
-    #     return self.payloads
 
     def show_sample_messages(self, count=5):
         """


docs/troubleshooting/syslog_pcap_sender.md
```diff
@@ -0,0 +1,120 @@
...
+## Overview
+
+`syslog_pcap_sender` is a universal Python script that extracts and replays syslog messages from PCAP files. It automatically 
+detects whether the capture contains TCP or UDP traffic and uses the appropriate method to send messages to a syslog server.
+Instead of replaying raw packets, `syslog_pcap_sender` extracts the syslog payloads and sends them over a fresh connection.
+
+## Features
+
+General:
+- Automatically detects the used protocol from PCAP and in case of both uses the one with more entries
```
Kawron: Maybe in case of two entries we could allow user to choose which one to use, or even better use two at the same time. Is it possible?

Kawron: Okay now I got to the end of documentation and I can see that you can force tcp or udp. So maybe to avoid confusing users, let's mention those option here as well, or just leave: "Automatically detects the used protocol from PCAP"

syslog_pcap_sender.py
```diff
@@ -0,0 +1,402 @@
...
+        pcap_file=args.input,
+        dest_ip=args.dest_ip,
+        dest_port=args.dest_port,
+        src_ip=args.src_ip,
+        src_port=args.src_port,
+        no_dedup=args.no_dedup
+    )
+
+    # Analyze PCAP
+    detected_protocol = sender.analyze_pcap()
```
Kawron: `.analyze_pcap` has a side effect of setting `sender.protocol()`. The name doesn't suggest that it does that and it has two responsibilities. If you don't know the `.analyze_pcap` body it looks like sb forgot to actually use `detected_protocol` variable (besides in the print statements).

I would change that to either:
```
    if args.force_tcp:
        print(f"\nForcing TCP mode (detected: {detected_protocol})")
        sender.protocol = 'TCP'
    elif args.force_udp:
        print(f"\nForcing UDP mode (detected: {detected_protocol})")
        sender.protocol = 'UDP'
     else:
         sender.protocol = detected_protocol
```

or create new method `set_protocol` and pass `force_tcp` `force_udp` arguments there

syslog_pcap_sender.py
```diff
@@ -0,0 +1,402 @@
...
+
+    # Override protocol if forced
+    if args.force_tcp:
+        print(f"\nForcing TCP mode (detected: {detected_protocol})")
```
Kawron: I think that the "(detected: " part is unneccessary as we already print this in `analyze_pcap()` and it will be printed right before

syslog_pcap_sender.py
```diff
@@ -0,0 +1,402 @@
...
+        for pkt in self.packets:
+            payload = None
+
+            # Extract payload based on protocol
+            if self.protocol == 'TCP' and TCP in pkt and Raw in pkt:
+                payload = bytes(pkt[Raw].load)
+                packets_with_payload += 1
+            elif self.protocol == 'UDP' and UDP in pkt and Raw in pkt:
+                payload = bytes(pkt[Raw].load)
+                packets_with_payload += 1
```
Kawron: maybe it will be cleaner to do a guard condition:

if !(self.protocol == 'TCP' and TCP in pkt and Raw in pkt or self.protocol == 'UDP' and UDP in pkt and Raw in pkt):
   continue
   
 Then you can move payload extraction and counter incrementing outside the if statements

syslog_pcap_sender.py
```diff
@@ -0,0 +1,402 @@
...
+
+            if payload and len(payload) > 0:
+                if self.no_dedup:
+                    # No deduplication - add all payloads
+                    self.payloads.append(payload)
+                else:
+                    # Deduplicate (important for TCP retransmissions)
+                    payload_hash = hash(payload)
```
Kawron: I think set in python already hashes payloads so it's unnecessary

syslog_pcap_sender.py
```diff
@@ -0,0 +1,402 @@
...
+
+        return self.payloads
+
+    def show_sample_messages(self, count=5):
+        """
+        Display sample messages from the capture
+        """
+        print(f"\nSample Messages (first {min(count, len(self.payloads))}):")
+
+        for i, payload in enumerate(self.payloads[:count]):
```
Kawron: enumrate accepts second argument which is number from which it starts counting so we can do:

`for i, payload in enumrate(self.payloads[:count], 1)`:

and then you don't have to add `i+1` later on

docs/troubleshooting/syslog_pcap_sender.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+### Install Script
+
+Download `syslog_pcap_sender` script from https://github.com/splunk/splunk-connect-for-syslog
+
+```bash
+
+# Make executable
+chmod +x syslog_pcap_sender
```
Kawron: missing .py

docs/troubleshooting/syslog_pcap_sender.md
```diff
@@ -0,0 +1,120 @@
...
+
+# Verify installation
+./syslog_pcap_sender --help
```
Kawron: missing .py

docs/troubleshooting/syslog_pcap_sender.md
```diff
@@ -0,0 +1,120 @@
...
+## Usage
+
+Preview Messages (Safe - No Sending)
+```bash
+./syslog_pcap_sender -i capture.pcap --extract-only -v
+```
+
+Send to Syslog Server
+```bash
+./syslog_pcap_sender -i capture.pcap -H 192.168.1.100 -P 514 -s 192.168.1.99
```
Kawron: wrong options, there is no -H it should be -d

syslog_pcap_sender.py
ajasnosz: Good point. I've moved setting the protocol to different method

syslog_pcap_sender.py
```diff
@@ -0,0 +1,675 @@
...
+                print(f"    - {packets_without_payload} UDP packets without payload")
+            if duplicate_count > 0 and not self.no_dedup:
+                print(f"    - {duplicate_count} duplicate payloads (deduplicated)")
+
+        if len(self.payloads) == 0:
+            print("No syslog payloads found in packets", file=sys.stderr)
+            sys.exit(1)
+
+        return self.payloads
+    # def _extract_udp_payloads(self):
```
Kawron: Undeleted commented code. Is there a reason for leaving it?

ajasnosz: removed it

PR #2896, opened 2026-01-27
Feat/functional tests for sc4s features

Kawron:


tests/feature-tests/test_features_config_generation.py
```diff
@@ -0,0 +1,384 @@
...
+    assert result.stdout.count('persist-name("DEFAULT_514_') == 2, \
+        f"Expected 2 UDP socket entries. Got:\n{result.stdout}"
+
+
+# =============================================================================
+# Receive Buffer Tests
+# =============================================================================
+
+@pytest.mark.features("rcv_buff")
+def test_udp_so_rcvbuf_enabled():
```
sbylica-splunk: Can we test SC4S_SOURCE_TCP_SO_RCVBUFF and SC4S_SOURCE_TLS_SO_RCVBUFF too? I'm pretty sure SC4S_SOURCE_TLS_SO_RCVBUFF is broken currently so the test should fail

Kawron: done

.github/workflows/ci-lite.yaml
```diff
@@ -411,6 +411,44 @@ jobs:
       - name: Run tests
         run: poetry run pytest tests/test_healthcheck_unit_tests.py
 
+  test-ebpf:
```
sbylica-splunk, line 414: Do we want to add this to: release -> needs (like other tests)?

tests/feature-tests/test_features_config_generation.py
```diff
@@ -0,0 +1,407 @@
+"""
```
sbylica-splunk, line 1: Looks like this test is failing on github

tests/feature-tests/test_features_config_generation.py
```diff
@@ -0,0 +1,407 @@
...
+# =============================================================================
+
+@pytest.mark.features("udp_iw")
+def test_udp_iw_use_enabled():
+    result = run_plugin({
+        "SC4S_SOURCE_UDP_IW_USE": "yes",
+    })
+
+    assert result.returncode == 0, f"Failed to generate config: {result.stderr}"
+    assert "log-iw-size(250000)" in result.stdout, \
```
sbylica-splunk, line 221: I'm not sure if we want to check for a particular value here, I'd rather only check if the variable exists

tests/feature-tests/test_features_config_generation.py
Kawron: That is a correct behaviour, because we have a bug regarding tls options

tests/feature-tests/test_features_config_generation.py
Kawron: I think it's better to check for a particular value. This way we are also testing the default value, and if someone will unintentionally change it, the test will fail

.github/workflows/ci-lite.yaml
Kawron: I think this a good idea, I will implement it

tests/feature-tests/test_features_config_generation.py
Kawron: Though, I can modify it. Let me know what you think

tests/feature-tests/test_features_config_generation.py
Kawron: Merged the tls PR so it works now

tests/feature-tests/test_features_config_generation.py
sbylica-splunk: I think it's up to you, I would prefer a less strict check but it's ok this way too

PR #2899, opened 2026-01-29
fix: use tls variables instead of tcp

Kawron:
Use tls config variables for tls section instead of using tcp variables

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -18,6 +18,9 @@ For a step by step guide [see here](./v3_upgrade.md).
 
 You may need to migrate legacy log paths or version 1 app-parsers for version 2. To do this, open an issue and attach the original configuration and a compressed pcap of sample data for testing. We will evaluate whether to include the source in an upcoming release.
 
+### Upgrade from <3.39.0
+Introduced new configuration options for TLS connections: `SC4S_SOURCE_TLS_MAX_CONNECTIONS`, `SC4S_SOURCE_TLS_IW_SIZE`, `SC4S_SOURCE_TLS_FETCH_LIMIT`, and `SC4S_SOURCE_LISTEN_TLS_SOCKETS`. Previously, TLS connections used TCP's configuration. The default values for TLS options are the same as their TCP counterparts.
+
 ### Upgrade from <3.37.0
 In `entrypoint.sh` the old variable mappings `SPLUNK_HEC_URL`, `SPLUNK_HEC_TOKEN`, `SC4S_DEST_SPLUNK_HEC_TLS_VERIFY` are deprecated and will not be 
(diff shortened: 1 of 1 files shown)


package/etc/conf.d/sources/source_syslog/plugin.jinja
sbylica-splunk, line 70: Looks like SC4S_SOURCE_LISTEN_TLS_SOCKETS should work, but is not described in the docs anywhere

sbylica-splunk: You can check these variables here:
https://github.com/splunk/splunk-connect-for-syslog/blob/d076b8e59c120272cf649a25263ba3aaed6b7aeb/package/enterprise/etc/conf.d/sources/source_syslog/plugin.py#L86

package/etc/conf.d/sources/source_syslog/plugin.jinja
```diff
@@ -78,11 +78,11 @@ source s_{{ port_id }} {
                 port({{ port }})
                 persist-name("{{ port_id }}_tls_{{ port }}_{{ i }}")
                 ip-protocol({{ enable_ipv6 }})
-                max-connections({{ port_tcp_max_connections }})
-                log-iw-size({{ port_tcp_log_iw_size }})
-                log-fetch-limit({{ port_tcp_log_fetch_limit }})
-                {%- if port_tcp_so_recvbuff >0 %}
-                so-rcvbuf({{ port_tcp_so_recvbuff }})
+                max-connections({{ port_tls_max_connections }})
```
sbylica-splunk, line 81: Same for: SC4S_SOURCE_TLS_MAX_CONNECTIONS,

package/etc/conf.d/sources/source_syslog/plugin.jinja
```diff
@@ -78,11 +78,11 @@ source s_{{ port_id }} {
...
+                log-iw-size({{ port_tls_log_iw_size }})
+                log-fetch-limit({{ port_tls_log_fetch_limit }})
```
sbylica-splunk, line 83: Both port_tls_log_iw_size and port_tls_log_fetch_limit are actually (...)_TCP_(...) anyway, we should probably change that

package/etc/conf.d/sources/source_syslog/plugin.jinja
Kawron: SC4S_SOURCE_LISTEN_TLS_SOCKETS added to documentation

package/etc/conf.d/sources/source_syslog/plugin.jinja
Kawron: Added to documentation

package/etc/conf.d/sources/source_syslog/plugin.jinja
Kawron: Done + added to documentation

ajasnosz (changes requested):
For backwards compatibility I think it would be better to set default as tcp env values.

PR #2909, opened 2026-02-04
Feat/parser for citrix version 14.1

Kawron:
With the new release of citrix 14.1 it is possible to enable rfc 5424 compliant body. Because it's not mandatory for citrix logs to follow this format and because we want to keep backward compability, I decided to create a new parser that will use tags assigned by custom port to match those events


package/etc/conf.d/conflib/netsource/app-netsource-citrix_netscaler_rfc5424.conf
```diff
@@ -0,0 +1,32 @@
...
+                product('netscaler')
+            );
+        };
+    };
+};
+
+application app-netsource-citrix_netscaler_rfc5424[sc4s-network-source] {
+    filter {
+        not filter(f_is_source_identified)
+        and tags(".source.s_CITRIX_NETSCALER_RFC5424")
```
sbylica-splunk, line 29: Do we need a custom port for that? In the issue we have this filter:
        ```match('^\d+-PPE-' value('PID'));```
 Can we use something similar instead?

PR #2935, opened 2026-03-02
fix: fix local sc4s and splunk setup for tests

Kawron:
This PR allows developers to run SC4S test suite, using docker compose. Previously, in order to run test locally and not in the CI/CD you had to first create SC4S and Splunk instance manually. Now you can just run the test command, and docker compose will automatically create necessary services. To achieve it, changes to pytest configuration were requried, the docker compose file had to be fixed (that means that you can use it to quickly create sc4s and splunk locally, even without running tests). Also, tests were adjusted to take account of the local timezone.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -138,6 +138,14 @@ def pytest_addoption(parser):
         default="latest",
         help="Splunk version",
     )
+    group.addoption(
+        "--splunk_docker_startup_wait",
+        action="store",
+        dest="splunk_docker_startup_wait",
+        type=int,
+        default=20,
+        help="Additional wait time in seconds after Splunk Docker is responsive",
+    )
 
 def is_responsive_splunk(splunk):
     try:
@@ -211,6 +219,13 @@ def start_splunk_docker(request, docker_services):
     docker_services.wait_until_responsive(
         timeout=180.0, pause=1.0, check=lambda: is_responsive_splunk(splunk)
     )
+    startup_wait = request.config.getoption("splunk_docker_startup_wait")
+    if startup_wait > 0:
+        logger.info(
+            "Splunk reported healthy; waiting %s seconds for full initialization...",
+            startup_wait,
+        )
+        sleep(startup_wait)
 
     return splunk
 
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -49,7 +49,7 @@ services:
       - SC4S_NETAPP_ONTAP_NEW_FORMAT=yes
       - SC4S_USE_VPS_CACHE=yes
   splunk:
-    image: splunk/splunk:10.0
+    image: splunk/splunk:latest
     hostname: splunk
     platform: linux/amd64
     ports:


tests/docker-compose.yml
```diff
@@ -7,113 +7,61 @@
...
+      - 514:514
+      - 601:601
+      - 5614:5514
+      - 5601:5601
+      - 6000:6000
+      - 6002:6002
+      - 8080:8080
+      - 9000:9000
+      - 9001:9001
+      - 9002:9002
```
sbylica-splunk, line 30: What is this change about?

tests/conftest.py
```diff
@@ -222,11 +316,18 @@ def sc4s_external(request):
...
+        request.fixturenames.append("sc4s_docker")
+        sc4s = request.getfixturevalue("sc4s_docker")
+    else:
+        raise ValueError
 
-@pytest.fixture()
-def setup_sc4s(sc4s):
-    return sc4s
-
+    yield sc4s
```
sbylica-splunk, line 345: Why yield?

Kawron: It is a leftover from old code, I renamed this function ad moved it lower in the file so github shows it as a new piece of the code but that was already here. 

I don't know why it was used here, maybe previosuly there was some cleanup logic after yield, that's common practice in pytest fixtures.  I can change this to return as this doesn't matter in this case

tests/docker-compose.yml
Kawron: If you are asking about format ( - port:port vs - "port") I think the new version is most common than the old one and more standard. Besides that I've modified the ports mapping to match the mapping we have in CI/CD.

However I noticed that I'm missing udp ports. I'm gonna fix this

tests/conftest.py
Kawron: Actually It stopped working after changing to `return`, so I will keep it as it was

tests/docker-compose.yml
```diff
@@ -7,113 +7,63 @@
...
-      - SC4S_DEST_SYSLOG_NCSYSLOG_PORT=2514
-      - SC4S_DEST_SYSLOG_NCSYSLOG_MODE=GLOBAL
-      - SC4S_DEST_SYSLOG_NCSYSLOG_IETF=no
-      - SC4S_DEST_BSD_NCBSD_HOST=ncbsd
-      - SC4S_DEST_BSD_NCBSD_PORT=2514
-      - SC4S_DEST_BSD_NCBSD_MODE=GLOBAL
-      # - SC4S_DEST_SPLUNK_HEC_SECOND_URL=https://splunk:8088
-      # - SC4S_DEST_SPLUNK_HEC_SECOND_TOKEN=${SPLUNK_HEC_TOKEN}
-      # - SC4S_DEST_SPLUNK_HEC_SECOND_TLS_VERIFY=no
+      - SC4S_DEST_SPLUNK_HEC_DEFAULT_HTTP_COMPRESSION=yes
```
ajasnosz, line 35: Were most of the env vars removed because they were set as default?

tests/docker-compose.yml
```diff
@@ -7,113 +7,63 @@
...
-      # - SC4S_DEST_SPLUNK_HEC_GLOBAL=yes
-      - SC4S_LISTEN_SIMPLE_ATA_SYSLOG_TCP_PORT=6500
-      - SC4S_LISTEN_SIMPLE_SURICATA_SYSLOG_TCP_PORT=6501
-      - SC4S_LISTEN_SIMPLE_BRO_SYSLOG_TCP_PORT=6502
-      - SC4S_LISTEN_DEFAULT_TCP_PORT=514,6504,6505,6506
-      # SC4S_USE_REVERSE_DNS=YES
-      # SC4S_REVERSE_DNS_KEEP_FQDN=YES
   splunk:
-    image: docker.io/splunk/splunk:latest
+    image: splunk/splunk:10.0
```
ajasnosz: do we want to stay with set Splunk version?

tests/docker-compose.yml
Kawron: Because this docker compose file didn't work at all when I started working on it, I assumed it as depricated. So I decided to use only the env variables that we use in our CI/CD pipiline, to make sure both of them are as similiar as they can be

tests/docker-compose.yml
Kawron: Good catch, I fill fix it

Kawron: Done, I've also added an additonal waiting time for splunk in docker, because I noticed that even if splunk reports itselfs healthy the tests can be flaky after building new container. This won't affect CI/CD pipeline

PR #2938, opened 2026-03-03
Config script updates

sbylica-splunk:


configuration-tool.sh
```diff
@@ -49,6 +98,74 @@ echo "This tool will help you generate an optimized SC4S configuration"
...
+            SPLUNK_URL="$input"
+            break
+        else
+            printf "${RED}Invalid URL format. Must start with http:// or https:// followed by a hostname.${NC}\n"
+            printf "${YELLOW}Example: https://splunk.example.com:8088${NC}\n"
+        fi
+    done
+}
+
+# Prompt for HEC token with validation, retries until valid; input is masked
```
Kawron: Comment states that input is masked, but when I tested it it's not

configuration-tool.sh
```diff
@@ -167,295 +389,205 @@ if [ "$mode_choice" = "2" ]; then
...
-echo "You can modify /etc/sysctl.conf following this documentation:"
-echo "https://splunk.github.io/splunk-connect-for-syslog/main/gettingstarted/getting-started-runtime-configuration/#tune-your-receive-buffer"
+    echo ""
+    printf "${YELLOW}Note: You need to adjust your system's receiving buffer to match the configured values.${NC}\n"
+    echo "Add the following to /etc/sysctl.conf:"
+    echo ""
+    if [ "$SC4S_SOURCE_UDP_SO_RCVBUFF" -gt 0 ]; then
+        printf "${GREEN}  net.core.rmem_default = %s${NC}\n" "$SC4S_SOURCE_UDP_SO_RCVBUFF"
+        printf "${GREEN}  net.core.rmem_max = %s${NC}\n" "$SC4S_SOURCE_UDP_SO_RCVBUFF"
+    elif [ "$SC4S_SOURCE_TCP_SO_RCVBUFF" -gt 0 ]; then
```
Kawron: Use two if's. I think it is possible for user to have both UDP and TCP rcv_buffers configured, which would print only UDP information

configuration-tool.sh
```diff
@@ -14,20 +58,25 @@ BLUE='\033[0;34m'
...
 SC4S_SOURCE_UDP_FETCH_LIMIT=1000
 SC4S_ENABLE_EBPF="no"
 SC4S_EBPF_NO_SOCKETS=4
 
 PARALLELIZE="no"
 SC4S_PARALLELIZE_NO_PARTITION=4
+SC4S_SOURCE_UDP_IW_USE="no"
+SC4S_SOURCE_UDP_IW_SIZE=1000000
 SC4S_SOURCE_TCP_IW_USE="no"
 SC4S_SOURCE_TCP_IW_SIZE=1000000
```
Kawron: Default value for TCP is 20000000

configuration-tool.sh
sbylica-splunk: Removed the comment

configuration-tool.sh
sbylica-splunk: Ok, changed both UDP and TCP to their default values

configuration-tool.sh
sbylica-splunk: Done

PR #2946, opened 2026-03-05
Attempte to override splunk host with deviceHostName for NETSCOUT

digrajpu:


tests/test_netscout_arbor-edge.py
```diff
@@ -46,3 +46,35 @@ def test_netscout_arboredge(record_property,  setup_splunk, setup_sc4s):
...
+    )
+    search = st.render(epoch=epoch, host=host)
+
+    result_count, _ = splunk_single(setup_splunk, search)
+
+    record_property("host", host)
+    record_property("resultCount", result_count)
+    record_property("message", message)
+
+    assert result_count == 1
```
Kawron, line 80: It doesn't test the new code change. It only checks if the event was indexed, but that is not what was changed in this PR. Modify the tests to check if the device hostname is properly extracted

digrajpu: `st = env.from_string(
        'search _time={{ epoch }} index=netids host="{{ host }}" sourcetype="netscout:aed"'
    )` 

I added host field in search criteria which matches the deviceHostName. Since the search only succeeds if the host field is correctly extracted and matched, this effectively validates the new extraction logic.

digrajpu:
Closing this PR as it no longer required changes regarding this issue pushed to new branch
https://github.com/splunk/splunk-connect-for-syslog/pull/2954

Issue #2700, opened 2025-02-11
Enable SNI configuration for SSL/TLS passthrough Loadbalancers


yoann-ls:
**What is the sc4s version?**
SC4S 3.34.2

**Feature Request description:**

SNI can improve load balancing of https traffic between SC4S and HECs  when it goes through a SSL/TLS passthrough loadbalancer. 
The sni() parameter is set by default to *no* in syslog-ng. 

Adding a configuration parameter in env_file to permit the configuration of  SNI in syslog-ng would solve this issue.

PR #2966, opened 2026-03-20
feat: add SNI configuration option for HEC TLS destinations

sbylica-splunk:
Add SC4S_DEST_SPLUNK_HEC_<ID>_SNI env var to enable TLS Server Name Indication on HEC destinations. This allows SSL/TLS passthrough load balancers to route HTTPS traffic based on the target hostname sent during the TLS handshake.

Made-with: Cursor


sbylica-splunk:
Pr addresses this issue: https://github.com/splunk/splunk-connect-for-syslog/issues/2700

PR #2970, opened 2026-03-23
Fix/dashboard version issue

digrajpu:


dashboard/dashboard.xml
```diff
@@ -209,7 +209,7 @@
       <title>SC4S version</title>
       <single>
         <search>
-          <query>| mcatalog values(sc4s_version) WHERE index=_metrics AND sc4s_container=$sc4s_instance$ | mvexpand values(sc4s_version)</query>
+          <query>| mcatalog values(sc4s_version) WHERE index=_metrics AND sc4s_container=$sc4s_instance$ | mvexpand values(sc4s_version) | sort - values(sc4s_version)</query>
```
sbylica-splunk: I think this is better but it's still possible that there will be an error, if the customer downgrades to the older sc4s version this will show the newer one instead.

If possible we should sort by Time instead and show the "latest (in time)" version used.

PR #2973, opened 2026-03-24
fix: fixed flaky dns resolver tests

sbylica-splunk:
- fix flaky reverse dns resolver tests
- fix flaky healthcheck tests
- fix deprecation warning: mem-buf-length()

### DIFF
--- a/tests/test_healthcheck_queue_size_limit.py
+++ b/tests/test_healthcheck_queue_size_limit.py
@@ -30,7 +30,7 @@ def main():
     print(f"Sending {args.limit} messages to {args.host}:{args.udp_port}...")
     send_messages(args.host, args.udp_port, "message", args.limit)
 
-    time.sleep(5)  # time to save the messages
+    time.sleep(1)  # time to save the messages
     print(f"Checking health status on {args.host}:{args.port}...")
     health_status = check_health(args.host, args.port)
 


configuration-tool.sh
```diff
@@ -86,7 +86,6 @@ SC4S_SOURCE_TCP_SO_RCVBUFF=-1
 SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_ENABLE="yes"
 SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_RELIABLE="no"
 SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_MEMBUFSIZE=10241024
-SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_MEMBUFLENGTH=15000
```
ajasnosz, line 89: are we backward compatible if someone will have this value set, would it throw error or crash?

tests/test_reverse_dns.py
```diff
@@ -27,30 +27,27 @@ def __setitem__(self, key, value):
...
     return socket.gethostbyname(domain)
 
-def get_host(ipaddr):
-    return socket.gethostbyaddr(ipaddr)
-
 @pytest.mark.addons("reverse-dns")
 def test_hostname_resolver_success():
     resolver = FixHostnameResolver()
     source_ip = get_ip_address("google.com")
-    resolved_host, _, _ = get_host(source_ip)
```
ajasnosz, line 37: if we remove the get_host isn't it just a dns test, no the reverse dns? What was the issue with this test before changes?

configuration-tool.sh
sbylica-splunk: It's an environmental value so it should just be ignored. I also think leaving a big deprecation note in release notes should help

tests/test_reverse_dns.py
sbylica-splunk: `resolved_host` is the "proper" hostname, but the problem is that IP can resolve to a different hostname randomly and for our google example it can be something like `pe-in-f139` vs `on-in-f139` or sth like that. This is why our test is flaky. 
We are testing resolver.parse which is an actual point of the test so it should be ok.
Instead of comparing to real hostname (which can be random) we are now just checking if it returns something that looks like a real hostname (is a string and has a dot inside). I think it works better this way

PR #2985, opened 2026-03-27
fix for queue size limit unit test

sbylica-splunk:


tests/test_healthcheck_unit_tests.py
```diff
@@ -197,6 +197,40 @@ def test_health_endpoint_no_queue_check(mock_run, client):
...
+@patch.dict(
+    os.environ,
+    {
+        "HEALTHCHECK_CHECK_QUEUE_SIZE": "true",
+    },
+    clear=True
+)
+@patch("package.sbin.healthcheck.check_queue_size", return_value=False)
+@patch("package.sbin.healthcheck.check_syslog_ng_health", return_value=True)
+def test_health_endpoint_queue_size_exceeded(mock_health, mock_queue, client):
```
sbylica-splunk, line 203: I don't think there is any way to make the original test not flaky, since we have to patch @patch("package.sbin.healthcheck.check_queue_size") for that. So I deleted it and made two unit tests instead.

PR #2994, opened 2026-03-30
fix: Extracting timezone from cisco_ios logs

digrajpu:


sbylica-splunk (commented):
Looks like there are some tests failing now

package/etc/conf.d/conflib/almost-syslog/app-almost-syslog-cisco_syslog.conf
```diff
@@ -106,16 +110,33 @@ block parser app-almost-syslog-cisco_syslog() {
...
                 '%b %d %I:%M:%S %p.%f',
                 '%b %d %I:%M:%S %p',
-                '%b %d %Y %I:%M:%S %p.%f'
+                '%b %d %Y %I:%M:%S %p.%f',
                 '%b %d %Y %H:%M:%S.%f',
                 '%b %d %Y %H:%M:%S',
                 '%Y-%m-%dT%T%z',
                 )
-                template("${.tmp.timestamp}"));
+                template("${.tmp.date_only}"));
```
sbylica-splunk: I think we should stay with tmp.timestamp and just add a (... %z) format, kind of like app-almost-syslog-citrix_netscaler.conf. I haven't tried it but will it work if you base parser here on the one from netscaler?

sbylica-splunk (changes requested):

package/etc/conf.d/conflib/almost-syslog/app-almost-syslog-cisco_syslog.conf
digrajpu: I already tried that, but it didn’t work because syslog-ng does not recognize text-based timezones.

sbylica-splunk: Well, we still need to do something different because this change breaks other parsers. Also there was a release so you should rebase on top of a new develop branch that was created after that.

tests/test_cisco_ios.py
```diff
@@ -73,6 +74,43 @@
 ]
 
 
+def return_timezone(message):
+    tz_list = ['CDT', 'DST', 'EDT', 'CEST', 'CET','PST','PDT','EST','BST','GMT','IST', 'JST']
```
Kawron, line 78: plz format the code. Use formatter like black or ruff

tests/test_cisco_ios.py
```diff
@@ -73,6 +74,43 @@
...
+        'BST': 'Europe/London',
+        'GMT': 'Europe/London',
+        'IST': 'Asia/Kolkata',
+        'JST': 'Asia/Tokyo',
+    }
+
+    # Get the standard IANA name from your map
+    iana_name = tz_map.get(abbr)
+    if not iana_name:
+        return epoch
```
Kawron, line 105: code smell: function is called `epoch_to_utc` but it still can return unchanged epoch. Throw error or at least return None and add typing

tests/test_cisco_ios.py
```diff
@@ -73,6 +74,43 @@
...
+
+    utc_dt = datetime.datetime.fromtimestamp(epoch, datetime.timezone.utc)
+    local_dt = utc_dt.replace(tzinfo=ZoneInfo(iana_name))
+    converted_utc_dt = local_dt.astimezone(datetime.timezone.utc)
+
+    return int(converted_utc_dt.timestamp())
```
Kawron, line 111: Why are we converting it to int? If you want to get rid of miliseconds, this is wrong approach. This will cause rounding errors, which will introduce hard to debug bugs.

tests/test_cisco_ios.py
```diff
@@ -102,13 +140,14 @@ def test_cisco_ios(
         host=host,
         year=year,
     )
-
+    tzname = return_timezone(message)
+    new_epoch = epoch_to_utc(int(epoch), tzname)
```
Kawron, line 144: again int(epoch) rounding errors

tests/test_cisco_ios.py
```diff
@@ -102,13 +140,14 @@ def test_cisco_ios(
         host=host,
         year=year,
     )
-
+    tzname = return_timezone(message)
```
Kawron, line 143: you are passing whole message to this function. What if in the log body there will be a string that matches the timezone abbr. Maybe pass only the stuff before `:`

tests/test_cisco_ios.py
```diff
@@ -245,13 +285,15 @@ def test_cisco_nx_os_soup2(
     message = mt.render(
         mark="<111>", bsd=bsd, host=host, date=date, time=time, tzoffset=tzoffset
     )
+    tzname = return_timezone(message)
+    new_epoch = epoch_to_utc(int(epoch), tzname)
```
Kawron, line 289: int(epoch)

Kawron, line 289: Won't UTC and GMT have always the same time?

tests/test_cisco_ios.py
```diff
@@ -73,6 +74,43 @@
 ]
 
 
+def return_timezone(message):
+    tz_list = ['CDT', 'DST', 'EDT', 'CEST', 'CET','PST','PDT','EST','BST','GMT','IST', 'JST']
+    return next((tz for tz in tz_list if tz in message), 'UTC')
+
+
+def epoch_to_utc(epoch, abbr):
```
Kawron, line 82: The naming is incorrect, that is not what this function does. It reads the epoch in utc (so why it's called epoch to utc when it reads the epoch as instant in utc?), changes the timezone while **keeping the wall-clock time** (this is important it changes the instant) and then coverts it back to utc and converts it to int.

So it takes epoch, interprets it like it's in utc, changes the time of this epoch, converts this new instant back to utc and converts it to int. The function name doesn't convey this behaviour

Kawron (changes requested):

package/etc/conf.d/conflib/almost-syslog/app-almost-syslog-cisco_syslog.conf
```diff
@@ -110,13 +110,28 @@ block parser app-almost-syslog-cisco_syslog() {
                 '%b %d %H:%M:%S',
                 '%b %d %I:%M:%S %p.%f',
                 '%b %d %I:%M:%S %p',
-                '%b %d %Y %I:%M:%S %p.%f'
+                '%b %d %Y %I:%M:%S %p.%f',
```
sbylica-splunk, line 113: Now I haven't tested it, but I "think" this is all that we need to fix here, since the date parser was broken because of a broken comma. Can you try it without all of the rewrite in the if block below and see if it works? I think those are superflous

sbylica-splunk (changes requested):

tests/test_cisco_ios.py
digrajpu: I don't want to remove the milliseconds or change the precision. I simply need the epoch time represented as an integer rather than a float.

tests/test_cisco_ios.py
digrajpu: Here the epoch was converted to an integer because datetime.datetime.fromtimestamp() expects a numeric timestamp.

tests/test_cisco_ios.py
digrajpu: Here the epoch was converted to an integer because datetime.datetime.fromtimestamp() expects a numeric timestamp.

tests/test_cisco_ios.py
Kawron: That's not true. From datetime documentation:

> Return the local date and time corresponding to the POSIX timestamp, such as is returned by [time.time()](https://docs.python.org/3/library/time.html#time.time). If optional argument tz is None or not specified, the timestamp is converted to the platform’s local date and time, and the returned [datetime](https://docs.python.org/3/library/datetime.html#datetime.datetime) object is naive.

and from time.time()

> Return the time in seconds since the [epoch](https://docs.python.org/3/library/time.html#epoch) as a floating-point number. The handling of [leap seconds](https://en.wikipedia.org/wiki/Leap_second) is platform dependent. On Windows and most Unix systems, the leap seconds are not counted towards the time in seconds since the [epoch](https://docs.python.org/3/library/time.html#epoch). This is commonly referred to as [Unix time](https://en.wikipedia.org/wiki/Unix_time).

And in the implementation you can see that it handles floats [_pydatetime.py](https://github.com/python/cpython/blob/f93834ff01128774532c101c574e47c6c0418540/Lib/_pydatetime.py#L1855):
```
    def _fromtimestamp(cls, t, utc, tz):
        """Construct a datetime from a POSIX timestamp (like time.time()).

        A timezone info object may be passed in as well.
        """
        frac, t = _math.modf(t)
        us = round(frac * 1e6)
        if us >= 1000000:
            t += 1
            us -= 1000000
        elif us < 0:
            t -= 1
            us += 1000000
```

tests/test_cisco_ios.py
Kawron: https://github.com/splunk/splunk-connect-for-syslog/pull/2994#discussion_r3123526753

tests/test_cisco_ios.py
Kawron: https://github.com/splunk/splunk-connect-for-syslog/pull/2994#discussion_r3123526753

package/etc/conf.d/conflib/almost-syslog/app-almost-syslog-cisco_syslog.conf
digrajpu: syslog-ng does not recognize text-based timezones so it not works

PR #3010, opened 2026-04-09
docs: add per-destination metadata override example

### LINKED ISSUE #2605: Different metadata values for every destination
```
SC4S version: 3.31.0
Host OS: Ubuntu 22.04.4 LTS (Jammy Jellyfish)
Runtime: Docker + systemd
```

I have to send data from my SC4S instance to a different Splunk Cloud deployment altoguether. The index names in this alternate destination differ from the ones I'm currently using. How can I modify the metadata that I send to the alternate destination so that the index names match without disturbing the metadata I send to my own Splunk Cloud deployment?

sbylica-splunk:
Documentation that can be used in similar situations as the one described in the issue:

Closes #2605 

Made-with: Cursor

### DIFF
--- a/docs/destinations.md
+++ b/docs/destinations.md
@@ -202,9 +202,9 @@ application sc4s-lp-cisco_dest_fmt_other[sc4s-lp-dest-select-d_hec_fmt_other] {
 };
 ```
 
-With this configuration, Cisco events are sent to both destinations: the default HEC uses the index from `splunk_metadata.csv`, while the OTHER HEC uses `cisco_alt_index`.
+With this configuration, Cisco events are sent to both destinations: the default HEC uses default index, while the OTHER HEC uses `cisco_alt_index`. Do mind that this index may still be overriden later (for example by metadata.csv) so it may be necessary to use a postfilter instead.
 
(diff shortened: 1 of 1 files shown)


docs/destinations.md
```diff
@@ -167,6 +167,45 @@ application sc4s-lp-cisco_ios_dest_fmt_other[sc4s-lp-dest-select-d_hec_fmt_other
...
+    filter {
+        'cisco' eq "${fields.sc4s_vendor}";
+    };
+    parser { sc4s-lp-cisco_dest_other_parser(); };
+};
+```
+
+With this configuration, Cisco events are sent to both destinations: the default HEC uses default index, while the OTHER HEC uses `cisco_alt_index`. Do mind that this index may still be overriden later (for example by metadata.csv) so it may be necessary to use a postfilter instead.
+
+**Note:** The index override in the selector applies only to the alternate destination. The default destination's metadata is not affected.
```
Kawron: We can consider using: `!!! note "Note"` syntax

Kawron (approved):

docs/destinations.md
sbylica-splunk: Yeah, looks better

PR #3017, opened 2026-04-13
fix:HTTP Proxy support for microk8s deployment

digrajpu:


charts/splunk-connect-for-syslog/templates/statefulset.yaml
```diff
@@ -53,6 +53,12 @@ spec:
                 fieldPath: spec.nodeName
           - name: SC4S_RUNTIME_ENV
             value: "k8s"            
+          {{- if .Values.sc4s.env }}
+          {{- range .Values.sc4s.env }}
+          - name: {{ .name }}
```
Kawron: Please provide more context on what those changes do. Moreover I don't see .Values.sc4s.env documented anywhere, and I don't see them in values.yaml

Kawron (changes requested):
Please provide more context

Kawron (changes requested):
Isn't it a little bit too general for this use case? In this variable we can set anything we want, not only http proxy. But the documentation and comments doesn't reflect that. In the `values.yaml` there is no mention that you can add other options as well, and in the k8s-microk8s.md doc there are no examples of using other options, it is just mentioned that you can use it for runtime settings. If we give users possibility to modify all environment variables used by sc4s we should documment it.
Moreover, what if I will pass env: `SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_ENABLE=yes`, it will be then overriden becuase it's explciitly set later in that file. That's also undocummented and can result in bugs.

Currently we lack feature that allows user to set any env variable in k8s distirbution, so in general this idea is not bad. But we would have to documment it properly, and change the way the current template is working e.g. don't set env that are already included unless there are explicitly set in values (see: `SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_ENABLE`). Or delete the old solution and just use sc4s.values.env.

Or we can do it like it's currently implemented and use set HTTP_PROXY explcitily. Again, see how it's done with `SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_ENABLE`.

@ajasnosz @sbylica-splunk What do you think about this?

charts/splunk-connect-for-syslog/values.yaml
```diff
@@ -13,6 +13,17 @@ splunk:
 
 sc4s:
   {}
+    env: []
```
sbylica-splunk: Pretty sure this won't parse, env should be inside the brackets

digrajpu:
Sure @Kawron To keep this PR focused and safe, I will change the implementation to use explicit variables (e.g., sc4s.proxy.http and sc4s.proxy.https) instead of a generic env block. This ensures clear documentation in values.yaml and avoids any accidental overrides of core settings. 
CC: @sbylica-splunk

ajasnosz:
I agree with @Kawron, and I think our k8s configuration is not full or wasn't updated for longer time, so we can add another task to cross check it and update in the future.

sbylica-splunk:
> I agree with @Kawron, and I think our k8s configuration is not full or wasn't updated for longer time, so we can add another task to cross check it and update in the future.

Honestly we could have a meeting at some point and go through kicrok8s configuration, I don't think it's very up to date and probably needs a makeover

docs/gettingstarted/k8s-microk8s.md
```diff
@@ -128,3 +128,19 @@ resources:
...
+# Proxy Configuration
+
+If your environment requires a proxy to reach external destinations you can configure the proxy settings in your values.yaml
+
+### Example Usage
+
+To enable proxy support, update your values.yaml as follows:
+
+```yaml
+sc4s:
```
Kawron, line 141: Shouldn't it be in the {}? In values.yaml we do have {}. But I'm not sure, can you confirm this?

Kawron (changes requested):

docs/gettingstarted/k8s-microk8s.md
digrajpu: I don’t think it needs to be in {}. I believe {} is used to represent an empty block.

Kawron (approved):

charts/splunk-connect-for-syslog/templates/statefulset.yaml
```diff
@@ -53,6 +53,19 @@ spec:
                 fieldPath: spec.nodeName
           - name: SC4S_RUNTIME_ENV
             value: "k8s"            
+          # Inject additional enviroment variable from values.yaml to add proxy support
+          {{- if .Values.sc4s.proxy.http }}
+          - name: HTTP_PROXY
```
sbylica-splunk, line 58: Pretty sure these should be lowercase, env variables are case sensitive here right?
https://splunk.github.io/splunk-connect-for-syslog/main/configuration/#configure-your-external-http-proxy

sbylica-splunk: Also when running:
 ```
helm template test . --set splunk.hec_url=test --set splunk.hec_token=foo
```
I get this:
```
Error: template: splunk-connect-for-syslog/templates/statefulset.yaml:57:24: executing "splunk-connect-for-syslog/templates/statefulset.yaml" at <.Values.sc4s.proxy.http>: nil pointer evaluating interface {}.http
```
Might be worth looking into, I will test this a bit more tomorrow

digrajpu: > Pretty sure these should be lowercase, env variables are case sensitive here right? https://splunk.github.io/splunk-connect-for-syslog/main/configuration/#configure-your-external-http-proxy

I tested this and it is working as expected let me know if it is not working at your end

Issue #2712, opened 2025-02-27
Thales SafeNet Trusted Access (STA) Support


Stjubit:
**What is the sc4s version?**

`3.30.1`

**Is there a pcap available? If so, would you prefer to attach it to this issue or send it to Splunk support?**

I have a packet capture, but it includes confidential customer data that I'm not allowed to share.

I anonymized an example log that you can use for testing though --> see attachment:

[thales_sta_example.log](https://github.com/user-attachments/files/19011491/thales_sta_example.log)

Please note that a BOM is in there and it's really important for making the parser work.

**What the vendor name?**

Thales

**What's the product name?**

SafeNet Trusted Access (STA)

**If you're requesting support for a new vendor, do you have any preferences regarding the default index and sourcetype for their events?**

Index: `netauth`
Sourcetype: `thales:sta:json`

**Do you have syslog documentation or a manual for that device??**

https://thalesdocs.com/sta/agents/logging/index.html

**Feature Request description:**

This topic came up in Slack first: https://splunkcommunity.slack.com/archives/CNV918JCQ/p1739203759977949.

It's one of the worst syslog messages you can get, because it says it's RFC-compliant and then sends invalid headers, which breaks default-parsers in sc4s and syslog-ng adds an error to the message (`Error processing log message`).

I managed to make it work by creating an own source with syslog-parser flag `no-parse` and manually extracting required fields using an app parser.

**Source:**

```
source s_THALES_STA {
  channel {
	source {
	  # NOTES:
	  # -> no-parse tells syslog-ng to not parse the message as RFC5424-compliant syslog,
	  #    which would fail because STA sends its' messages saying they are RFC compliant,
	  #    but actually they are not, which breaks further parsing.
	  # -> no-multi-line is important, because otherwise full message can't be rewritten
	  syslog (
		transport("tcp")
		so-reuseport(1)
		port(602)
		persist-name("THALES_STA_6587_602_1")
		ip-protocol(4)
		max-connections(2000)
		log-iw-size(20000000)
		log-fetch-limit(2000)
		keep-timestamp(yes)
		use-dns(no)
		use-fqdn(no)
		chain-hostnames(off)
		flags(validate-utf8, no-parse, no-multi-line)
	  );
	};

	parser {
	  channel {
		rewrite {
		  set("Thales", value(".netsource.sc4s_vendor"));
		  set("STA", value(".netsource.sc4s_product"));
[...]

PR #3036, opened 2026-04-20
feat: support for thales sta

Kawron:
PR for https://github.com/splunk/splunk-connect-for-syslog/issues/2712.

This is a fix for non-compliant logs send by Thales. Those logs are octet-framed but the syslog header is malformed, because of that we have to use no-parse flag, syslog() driver and we cannot use syslog parsers that try to parse the header. Because of that I've added a new listener that does all that and directly send the log to netsource topic. So the header extraction has to be done directly by netsource parser.

This does not work for any source path because of two issues: header format is incorrect, mainly it puts BOM it sdata field, and it uses new lines in json message body. Becuase of that when sent to port 514:
image
and when sent to port 601:
image


tests/test_thales_sta.py
```diff
@@ -0,0 +1,94 @@
...
+    '    "resultText": "AUTH_SUCCESS",\n'
+    '    "agentId": "6",\n'
+    '    "message": "",\n'
+    '    "usedName": "XXXXX",\n'
+    '    "credentialType": "GrIDsure"\n'
+    "  }\n"
+    "}"
+)
+
+
```
ajasnosz, line 44: It would be good to add here the commended line with unedited example of the logs, as we  do in most of the tests

ajasnosz (changes requested):

tests/test_thales_sta.py
Kawron: done

ajasnosz (approved):

.github/workflows/ci-lite.yaml
```diff
@@ -226,6 +226,7 @@ jobs:
           SC4S_LISTEN_CISCO_ESA_TCP_PORT: 9000
           SC4S_LISTEN_RARITAN_DSX_TCP_PORT: 9001
           SC4S_LISTEN_CITRIX_NETSCALER_RFC5424_TCP_PORT: 9002
+          SC4S_LISTEN_THALES_STA_RFC6587_NOPARSE_PORT: 9003
```
sbylica-splunk, line 229: Ok now looking at it, makes total sense, but would it be possible to create a more "generic" configuration key (like SC4S_RFC6587_NOPARSE_PORT for example) and make it use the no-parse flag, but in a way that allows using this port for other products later?
So a generic port that is later used by thales, with a possibility for other parsers being used for other things?

Kawron: Hmmmm, I would have to look at it. Maybe it is possible it depends on how much the tcp and udp configuration differs. But yes, generally that is sth I wanted to achieve, what I did is already some generalization but for tcp only, I though that we could just duplicate it for udp, but if this can be done in single if statment, it would be much better.

sbylica-splunk (approved):

PR #3038, opened 2026-04-20
Merge changes from main to develop

sbylica-splunk:


Kawron (approved):
Hmmm, I'm not sure what will be the state of the develop after that and if this won't cause any merge issues when trying to release next time. But we can approve it and in worst case just revert. Just in case, before merging this change clone the repo, so you can alway push --force from that copy

PR #3044, opened 2026-04-24
feat: added warnings for some security config validator options

sbylica-splunk:


configuration-tool.sh
```diff
@@ -630,3 +630,22 @@ printf "${YELLOW}Note: Enabling eBPF may require additional system permissions.$
...
+    echo "SC4S will not validate the Splunk HEC server's certificate, which exposes the connection"
+    echo "to attacks. This setting is intended for development/testing with self-signed"
+    echo "certificates. For production, use a certificate signed by a trusted CA and re-enable TLS"
+    echo "verification."
+    echo "Documentation: https://splunk.github.io/splunk-connect-for-syslog/main/configuration/#configure-your-syslog-source-tls-certificate"
+fi
+
+if [[ "$SPLUNK_URL" =~ ^http:// ]]; then
+    echo ""
+    printf "${YELLOW}Warning: Splunk HEC URL uses plaintext HTTP instead of HTTPS.${NC}\n"
```
Kawron: I would change a wording here a little bit. It reads like a statment that Splunk HEC URL in *general* doesn't use https. I would rewrite to sth like this:
Warning: Provided Splunk HEC URL uses plaintext HTTP instead of HTTPS...

Kawron (changes requested):

Kawron (approved):

Issue #3041, opened 2026-04-22
app-syslog-netapp_ontap.conf is missing supported timezone formats


ehlo550:
**What is the sc4s version ?**
3.37.0

**Is there a pcap available? If so, would you prefer to attach it to this issue or send it to Splunk support?**
```
<14>2026-04-22T08:54:42Z IBKNCL01-b1: IBKNCL01-b1: 00000061.03ba10d1 08297752 Wed Apr 22 2026 10:54:39 +02:00 [kern_audit:info:3768] 8503e80000a16343 :: IBKNCL01:http :: 172.24.81.160:33964 :: IBKNCL01:harvest2:harvest-rest-role :: GET /api/private/cli/aggr?fields=aggregate%2Ccomposite%2Cnode%2Cstorage_type%2Cuses_shared_disks&max_records=500&return_records=true :: Success: 

<14>Apr 22 10:55:01 IBKNCL01-b1: IBKNCL01-b1: 00000061.03ba112f 08297812 Wed Apr 22 2026 10:55:00 +02:00 [kern_audit:info:3768] 8503e80000a16375 :: IBKNCL01:http :: 172.24.84.74:52282 :: IBKNCL01:admin:admin :: GET /api/private/cli/vserver/fpolicy/engine/?fields=server-status,disconnect-reason,server-type&max_records=100 :: Pending 
```

**Is the issue related to the environment of the customer or Software related issue?**
both

**Describe the bug**
Current versions of netapp seem to emit logs with different timezone formats than the parser allows.
this leads to
```
dtparse: Expected: %a %d %Y %H:%M:%S %z; Actual: Apr 22 2026 10:55:21 +02:00
```

This is what I can configure on the device: 
```
-timestamp-format-override ?        
  no-override                 Default timestamp format based on the message format (rfc-3164 if message format is legacy-netapp,
                              iso-8601-local-time if message format is rfc-5424)
  rfc-3164                    RFC-3164 (format: Mmm dd hh:mm:ss)
  iso-8601-utc                ISO-8601 in UTC (format: YYYY-MM-DDThh:mm:ssZ)
  iso-8601-local-time         ISO-8601 in local time (format: YYYY-MM-DDThh:mm:ss+/-hh:mm)
```

[...]

PR #3046, opened 2026-04-24
fix: netapp date parsing

sbylica-splunk:
Fix for: https://github.com/splunk/splunk-connect-for-syslog/issues/3041

### DIFF
--- a/package/etc/conf.d/conflib/syslog/app-syslog-netapp_ontap.conf
+++ b/package/etc/conf.d/conflib/syslog/app-syslog-netapp_ontap.conf
@@ -7,7 +7,7 @@ block parser app-syslog-netapp_ontap() {
             );
             date-parser-nofilter(
                 format(
-                    '%a %d %Y %H:%M:%S %z',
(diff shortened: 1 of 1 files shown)


package/etc/conf.d/conflib/syslog/app-syslog-netapp_ontap.conf
```diff
@@ -7,7 +7,7 @@ block parser app-syslog-netapp_ontap() {
             );
             date-parser-nofilter(
                 format(
-                    '%a %d %Y %H:%M:%S %z',
+                    '%b %d %Y %H:%M:%S %z',
```
sbylica-splunk, line 10: The same pattern can be found in:
https://github.com/splunk/splunk-connect-for-syslog/blob/a4f6210375bd0d97a078cf1ee79af14399be5459/package/lite/etc/addons/netapp/app-netsource-netapp_ontap.conf

Looks like it was omitted here. Anyway, this is a slightly different fix than suggested, but it works and there are no errors after implementing it.

Kawron (approved):

PR #3047, opened 2026-04-24
fix: timezone extraction

digrajpu:


package/etc/conf.d/conflib/almost-syslog/app-almost-syslog-cisco_syslog.conf
```diff
@@ -110,13 +110,28 @@ block parser app-almost-syslog-cisco_syslog() {
                 '%b %d %H:%M:%S',
                 '%b %d %I:%M:%S %p.%f',
                 '%b %d %I:%M:%S %p',
-                '%b %d %Y %I:%M:%S %p.%f'
+                '%b %d %Y %I:%M:%S %p.%f',
```
sbylica-splunk, line 113: Going back to my question from previous pr: https://github.com/splunk/splunk-connect-for-syslog/pull/2994#discussion_r3145011088

Are we sure? How does the event look like in splunk if you change it this way?

digrajpu: "Apr 27 2026 04:35:12 PM.123456" this is the string it will match with

package/etc/conf.d/conflib/almost-syslog/app-almost-syslog-cisco_syslog.conf
```diff
@@ -110,13 +110,28 @@ block parser app-almost-syslog-cisco_syslog() {
...
                 '%b %d %Y %H:%M:%S.%f',
                 '%b %d %Y %H:%M:%S',
                 '%Y-%m-%dT%T%z',
                 )
                 template("${.tmp.timestamp}"));
             };
+            if (match("CEST|CET" value(".tmp.tz"))) {
```
Kawron: This implementation can cause bugs. Timezones: CEST, CET, PST PDT etc. always represent the same offset relative to the GMT. But the IANA zones e.g. Europe/Berlin can change depending on the daylight saving time. `fix-time-zone` keeps the wall-clock time and changes the timezone. So this code will behaviour differently in summer than in winter. Also what is the criteria for choosing timezones here? Clearly that's not all of them.

Kawron: I think you can pass to fix-time-zone the offset directly e.g. `fix-time-zone("+01:00")` but I'm not sure about that

Kawron (changes requested):

package/etc/conf.d/conflib/almost-syslog/app-almost-syslog-cisco_syslog.conf
sbylica-splunk: Ok, tested it and it seems fine, what about @Kawron comments?

package/etc/conf.d/conflib/almost-syslog/app-almost-syslog-cisco_syslog.conf
digrajpu: > This implementation can cause bugs. Timezones: CEST, CET, PST PDT etc. always represent the same offset relative to the GMT. But the IANA zones e.g. Europe/Berlin can change depending on the daylight saving time. `fix-time-zone` keeps the wall-clock time and changes the timezone. So this code will behaviour differently in summer than in winter. Also what is the criteria for choosing timezones here? Clearly that's not all of them.

I selected these specific timezones based on the current test input logs to cover our existing use cases. Please let me know if there are any additional timezones you would like me to include, and I will be happy to add them.

> I think you can pass to fix-time-zone the offset directly e.g. `fix-time-zone("+01:00")` but I'm not sure about that

I originally considered passing the offset directly, but I pivoted to the current implementation to make the logic more dynamic. Using specific timezone identifiers instead of hardcoded offsets helps us better handle edge cases like Daylight Saving Time (DST).

digrajpu: What about this log it contains DST timezone it will change region to region like In New York, DST is -04:00. In London, it is +01:00. In Sydney, it is +11:00.

`"{{ mark }}84027: {{ bsd }}.{{ millisec }} DST: %SYS-5-CONFIG_I: Configured from console by username on vty0 ({{ host }})",`

And What if IST timezone comes in logs it refers to three different timezones like 

- Indian Standard Time UTC+5:30, 
- Irish Standard Time: UTC+1.
- Israel Standard Time: UTC+2

Kawron (approved):

package/etc/conf.d/conflib/_common/tz_abbrev.csv
```diff
@@ -0,0 +1,29 @@
+GMT,.tmp.tz_fix,Etc/GMT
```
ajasnosz, line 1: can we put all standard timezone names here and sort the file by gmt+/- ?

ajasnosz (changes requested):

package/etc/conf.d/conflib/_common/tz_abbrev.csv
Kawron: There is no standard for this format. The whole thing predates the internet and is generally discouraged to use it. Most importantly, some of them are ambiguous, e.g. CST can mean Chine time, Usa Time or Cuba Time. This list was based on cisco documentation: https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/system-management/b-timing-and-synchronization.pdf

Kawron: Technically we could add more, but if cisco devices won't use them I think there is no point to it

ajasnosz (approved):

PR #3048, opened 2026-04-28
fix: http proxy support

digrajpu:


Kawron (approved):

charts/splunk-connect-for-syslog/templates/statefulset.yaml
```diff
@@ -53,6 +53,19 @@ spec:
                 fieldPath: spec.nodeName
           - name: SC4S_RUNTIME_ENV
             value: "k8s"            
+          # Inject additional enviroment variable from values.yaml to add proxy support
+          {{- if .Values.sc4s.proxy.http }}
+          - name: HTTP_PROXY
```
sbylica-splunk: Ok, seems to work well!
One thing we could check: https://splunk.github.io/splunk-connect-for-syslog/main/configuration/#configure-your-external-http-proxy

Would it be possible for these values to be lowercase? To keep it consistent with our previous use.

sbylica-splunk (approved):

Kawron (approved):

PR #3051, opened 2026-04-29
fix: app-syslog-f5_bigip not properly covered by the tests

digrajpu:


package/etc/conf.d/conflib/syslog/app-syslog-f5_bigip.conf
```diff
@@ -66,6 +66,12 @@ block parser app-syslog-f5_bigip() {
                 or program('apmd' type(string) flags(prefix))
                 or program('sshd(pam_audit)' type(string) flags(prefix))
                 or message('pam_radius_auth|authenticated|pam_bigip_authz|mod_auth_pam')
+                or program('audit_forwarder' type(string) flags(prefix))
```
Kawron: Why this was added to the parser? In general we avoid adding filters for program that are to general and could match others vednors (https://splunk.github.io/splunk-connect-for-syslog/main/sources/). And this task shouldn't require changing parsers.

It is not entirely wrong that some tests use `test-f5` prefix hostname if there is no other way of matching them, but those cases are sometimes mixed in with logs that should be catched by program filters on their own. E.g. 

```
testdata_f5bigip_syslog = [
    '{{ mark }}{{ bsd }} {{ host }} notice sshd(pam_audit)[27425]: user=root(root) partition=[All] level=Administrator tty=ssh host=192.168.2.100 attempts=1 start="Mon Dec 22 18:40:19 2014" end="Mon Dec 22 18:45:50 2014".',
    "{{ mark }}{{ bsd }} {{ host }} notice httpd[16784]: pam_bigip_authz: authenticated user user23 with role 0 (Administrator) in partition [All]",
    "{{ mark }}{{ bsd }} {{ host }} notice sshd[20797]: pam_radius_auth: pam_radius_auth: user user15 successfully authenticated",
    "{{ mark }}{{ bsd }} {{ host }} notice httpd[16784]: pam_bigip_authz: authenticated user user23 with role 0 (Administrator) in partition [All]",
    '{{ mark }}{{ bsd }} slot1/{{ host }} notice httpd[31905]: 01070417:5: AUDIT - user admin - RAW: httpd(mod_auth_pam): user=admin(admin) partition=[All] level=Administrator tty=/sbin/nologin host=192.168.4.39 attempts=1 start="Mon Dec 22 14:04:15 2014" end="Mon Dec 22 15:20:20 2014".',
]
```

Kawron (changes requested):
There is no need of changing the parsers

package/etc/conf.d/conflib/syslog/app-syslog-f5_bigip.conf
digrajpu: noted

Kawron (approved):

ajasnosz (approved):

PR #3065, opened 2026-05-13
feat: mcp authentication

Kawron:
This PR implements support for TLS and token based authentication for **both** SC4S API and MCP Server.

Both aps have different implementation though they are quite simillar. So I recommend reviewing first package/sbin and then mcp_server folders. I decided not to write a single fit all implementation because:
1. Those are two different aps, that can are independent of each other and can are deployed separately. If we would like to have shared implementation they would have to import each other
2. The are small differences between those two, e.g. mcp server token authentication extends FastMcp class for authentication and they use two different web servers: uvicorn vs gunicorn
3. Keeping implementation seperate give us more flexibility
4. It should be fairly easy to extract some of the methods to common util package that will be imported by both aps, if the cost of maintaince increases


package/sbin/api.py
```diff
@@ -19,6 +24,46 @@
...
+    if not tokenVerifier.verify_token(presented):
+        logger.warning(
+            "Authentication failed: %s %s from %s",
+            request.method,
+            request.path,
+            request.remote_addr,
+        )
+        return jsonify({"error": "Authentication failed"}), 401
+
+
```
sbylica-splunk, line 70: By default the api server should not be active, only available after user turns it on

Kawron: Done, the SC4S managment API is now enabled by `SC4S_API_MANAGEMENT_ENABLED` by default set to false

tests/test_management_api_auth.py
```diff
@@ -0,0 +1,106 @@
+"""Unit tests for package/sbin/auth.py (management API bearer-token auth)."""
+
+import pytest
+
+from auth import (
+    AUTH_TOKEN_ENV,
```
sbylica-splunk: This one doesn't exist

sc4s_mcp/health_probe.py
```diff
@@ -0,0 +1,32 @@
...
+
+import os
+import ssl
+import sys
+import urllib.request
+
+import tls
+from server import DEFAULT_PORT, HEALTH_PATH
+
+_TIMEOUT_SEC = 3
```
sbylica-splunk: This seems pretty low

docs/configuration.md
```diff
@@ -10,6 +10,116 @@ SC4S is primarily controlled by environment variables. This topic describes the
...
 
+## SC4S management API endpoints
+
+By default the management REST API exposes only the `/health` endpoint. The configuration and metadata endpoints (`/config/env`, `/config/parser`, `/config/parser/<name>`, `/config/parsers`, `/config/metadata/splunk`, `/config/metadata/compliance`) are **disabled** and return HTTP 404 unless explicitly opted in.
+
+| Variable | Values | Description |
+|----------|--------|-------------|
+| `SC4S_API_MANAGEMENT_ENABLED` | `true`/`1`/`yes`/`y`/`t` to enable; unset or any other value to disable (default) | When unset, only `/health` is registered. All `/config/*` paths return HTTP 404. |
+
+When you enable management endpoints, consider configuring `SC4S_AUTH_TOKEN_FILE` as well – the management endpoints can read and modify the syslog-ng configuration.
```
sbylica-splunk: This should probably be written in capital letters or sth. I would even consider disabling api entirely if there is no token, seems like a serious security risk

package/sbin/api.py
```diff
@@ -1,23 +1,82 @@
...
+    if not auth_header:
+        return jsonify({"error": "Missing Authorization header"}), 401
+
+    parts = auth_header.split()
+
+    if len(parts) != 2 or parts[0].lower() != "bearer":
+        return jsonify(
+            {"error": "Invalid Authorization format. Expected: Bearer <token>"}
+        ), 401
```
sbylica-splunk: There should be only one generic error here, this reveals too much to a potential attacker.

https://owasp.org/www-community/Improper_Error_Handling#
https://cheatsheetseries.owasp.org/cheatsheets/Error_Handling_Cheat_Sheet.html

package/sbin/api.py
```diff
@@ -1,23 +1,82 @@
...
+        "Management API endpoints enabled (%s=true): /config/* routes registered",
+        MANAGEMENT_ENABLED_ENV,
+    )
+else:
+    logger.info(
+        "Management API endpoints disabled (%s unset/false); only /health is exposed",
+        MANAGEMENT_ENABLED_ENV,
+    )
+
+tokenVerifier = build_token_verify()
```
sbylica-splunk: token_verifier

sbylica-splunk (changes requested):

sc4s_mcp/utils/http.py
```diff
@@ -1,36 +1,64 @@
...
     """Execute an HTTP request against the SC4S API with unified error handling."""
     url = f"{SC4S_API_URL}{path}"
+    headers = {**_auth_headers(), **kwargs.pop("headers", {})}
+    if headers:
+        kwargs["headers"] = headers
+    kwargs.setdefault("verify", _verify())
+    logger.debug("SC4S API %s %s", method.upper(), url)
     try:
-        resp = _METHODS[method](url, **kwargs)
+        resp = getattr(httpx, method)(url, **kwargs)
```
digrajpu, line 39: What is the default timeout for HTTP requests? If one is not configured explicitly, requests may hang indefinitely depending on network/service conditions.

tests/test_management_api_auth.py
Kawron: fixed

sc4s_mcp/health_probe.py
Kawron: increased to 30

sc4s_mcp/utils/http.py
Kawron: valid point, default is 5 second but I made it explicit

package/sbin/api.py
Kawron: done

docs/mcp_server/installation.md
```diff
@@ -19,19 +18,27 @@ SC4S MCP server is currently available only for Podman or Docker runtimes.
...
 |---|---|---|
-| `MCP_TRANSPORT` | `stdio` | Transport mode: `stdio` for local clients, `sse` for remote clients. |
-| `MCP_HOST` | `0.0.0.0` | Bind address used in `sse` mode. |
-| `MCP_PORT` | `8000` | TCP port used in `sse` mode. |
+| `MCP_TRANSPORT` | `stdio` / `http` | Transport mode: `stdio` for local clients, `http` for remote clients. |
+| `MCP_HOST` | `0.0.0.0` | Bind address used in `http` mode. |
+| `MCP_PORT` | `8000` | TCP port used in `http` mode. |
+| `MCP_LOG_LEVEL` | `INFO` | Logging verbosity. Accepts standard Python log level names: `DEBUG`, `INFO`, `WARNING`, `ERROR`. |
 | `SC4S_API_URL` | `http://localhost:8080` | URL of the SC4S management REST API. The MCP server calls this URL for all management tools. |
+| `SC4S_API_TOKEN` | _unset_ (no auth) | Bearer token sent by the MCP server to the SC4S management REST API in `Authorization: Bearer <token>`. Required when the SC4S API has authentication enabled. See [SC4S API authentication](#sc4s-api-authentication-optional) and [Enabling auth on the SC4S API](../configuration.md#sc4s-management-api-authentication). |
```
ajasnosz: If it's empty I would just leave default filed as empty instead of writing _unset_ . Similar for fields below

docs/mcp_server/installation.md
```diff
@@ -103,6 +110,152 @@ the container status with:
...
+
+```bash
+mkdir -p /opt/sc4s-mcp/tls
+openssl req -x509 -newkey rsa:2048 -nodes -days 365 \
+  -subj "/CN=<MCP_HOST>" \
+  -addext "subjectAltName=DNS:<MCP_HOST>" \
+  -keyout /opt/sc4s-mcp/tls/server.key \
+  -out    /opt/sc4s-mcp/tls/server.crt
+chmod 600 /opt/sc4s-mcp/tls/server.key
+```
```
ajasnosz: I"m not sure if we should suggest that. Or if we want to keep to for ourselves I wouldn't kept it in public docs, we can have a page that is not listed in the mkdocs, or have it on confluence blog etc

package/sbin/api.py
Kawron: done

docs/configuration.md
```diff
@@ -10,6 +10,116 @@ SC4S is primarily controlled by environment variables. This topic describes the
...
+
+## SC4S management API authentication
+
+The SC4S management REST API (default port `8080`) supports optional bearer-token authentication. When `SC4S_API_MANAGEMENT_ENABLED` is set and `SC4S_AUTH_TOKEN_FILE` is unset or empty, the management endpoints are accessible without credentials. When it is set, every request must carry a matching `Authorization: Bearer <token>` header; mismatches return HTTP 401.
+
+| Variable | Values | Description |
+|----------|--------|-------------|
+| `SC4S_AUTH_TOKEN_FILE` | _unset_ (auth disabled) | Path to a file containing the bearer token. When set, enables bearer-token authentication on the management REST API. |
```
ajasnosz: same here

docs/configuration.md
```diff
@@ -10,6 +10,116 @@ SC4S is primarily controlled by environment variables. This topic describes the
...
+The MCP server's configuration and metadata tools also require `SC4S_API_MANAGEMENT_ENABLED=true` on the SC4S side; without it, those tools will receive HTTP 404 responses from the SC4S API.
+
+## SC4S management API TLS
+
+The management API supports optional TLS. When enabled, the API serves
+HTTPS only — plaintext HTTP connections to the same port are refused.
+TLS is disabled by default.
+
+| Variable | Values | Description |
+|----------|--------|-------------|
```
ajasnosz: The values can be changed to Default, but if every one of them is unset/empty maybe we don't need to specify that in separate column?

sc4s_mcp/utils/token_utils.py
```diff
@@ -0,0 +1,22 @@
+"""Shared helper for loading bearer tokens from env var or secret file."""
+
+import os
+
+
+def load_token(token_env: str, token_file_env: str) -> str:
```
ajasnosz, line 6: This is pretty generic read/load from file function, I see similar in sbin/auth.py, I know they aren't ship together, but could you cross check the code if we don't have similar method used in the same modules so we don't duplicate the code and make it more generic like read from file functions?

ajasnosz (changes requested):

docs/configuration.md
Kawron: I would add this as a note. There is whole section about Authentication right below. I would keep auth optional. During design process this was discussed and we've agreed that the authentication should be kept simple and opt-in as in most cases SC4S would be deployed in private network. But we can discuss it again during SC4S sync.

sc4s_mcp/utils/token_utils.py
Kawron: Double checked, it is not duplicated in the same module

docs/mcp_server/installation.md
Kawron: Hmm good point I will remove it from here and then later prepare an confluence page. I also have some notes that are unfit to the official docs so I can add them there.

ajasnosz (approved):

package/sbin/source_ports_validator.py
```diff
@@ -14,23 +14,30 @@ def validate_source_ports(sources: list[str]) -> None:
...
-    for source, proto in source_ports.keys():        
+    for source, proto in source_ports.keys():
         for port in source_ports[(source, proto)]:
             if not port or port == "disabled":
                 continue
 
             elif not is_valid_port(port):
-                logger.error(f"SC4S_LISTEN_{source}_{proto}_PORT: {port} must be integer within the range (0, 65565)")
+                logger.error(
+                    f"SC4S_LISTEN_{source}_{proto}_PORT: {port} must be integer within the range (0, 65565)"
```
sbylica-splunk: Should be 65535

sbylica-splunk (approved):

digrajpu (approved):

PR #3067, opened 2026-05-14
feat: Adding support to dell OME product

digrajpu:


sbylica-splunk:
We should also add a doc in "known vendors"

package/lite/etc/addons/dell/app-syslog-dell_ome.conf
```diff
@@ -0,0 +1,48 @@
...
+            set("${idrac_ip}", value("fields.idrac_ip"));
+            set("${idrac_severity}", value("fields.idrac_severity"));
+            set("${idrac_messageid}", value("fields.idrac_messageid"));
+        };
+    };
+};
+
+application app-syslog-dell_ome[sc4s-syslog] {
+    filter {
+        message('{ IP }' type(string) flags(substring)) and
```
sbylica-splunk: I feel like these are too generic, do we have any documentation on how these events are structured?
Looking at the logs attached to the issue, I think at least "EEMI Audit event" would be better, but we probably need a good look at docs to improve this part

sbylica-splunk (changes requested):

package/lite/etc/addons/dell/app-syslog-dell_ome.conf
digrajpu: I reviewed the documentation and found that Dell OME does not provide a complete syslog schema or defined field structure for its events. Based on logs, we can apply filtering using conditions such as:

`message('OME' type(string) flags(substring)) or
message('EEMI' type(string) flags(substring)) or
message('Audit Event' type(string) flags(substring))`

Kawron: I think that having `{ <field> }` is quite unique, especially that those are connected with conjuction. If this is not a standard but a a thing specific to Dell OME, I think we can allow it. 

The question is do we have any other Dell sources that may also use this format?

Kawron: If we want to be even more precise we can do a regex, it will be slower match but we can then enforce that all of those three substring occur and in what order

Kawron: Okay, after inspecting the logs I think that overall `EEMI Audit event` is better as this is the only type of logs that customer provided

digrajpu: There is one more log with `EEMI System Health event` is present

Kawron: Then I think the best solution would be a regex: EEMI .+ { IP } .+ {HOSTNAME} and so on. You can add additional fields if you can see that they repeat in each log

Kawron (changes requested):
I would implement it as a regex (see creating filters doc page) that would enforce the order of those matching fields. Now something like this would match
`{ Severity }{ HostName}.  {MessageID} EEMI { IP }`

package/lite/etc/addons/dell/app-syslog-dell_ome.conf
```diff
@@ -0,0 +1,48 @@
...
+            set("${idrac_severity}", value("fields.idrac_severity"));
+            set("${idrac_messageid}", value("fields.idrac_messageid"));
+        };
+    };
+};
+
+application app-syslog-dell_ome[sc4s-syslog] {
+    filter {
+          match(
+            'EEMI.*\{\s*IP\s*\}.*\{\s*HostName\s*\}.*\{\s*Severity\s*\}.*\{\s*MessageID\s*\}.*' value("MESSAGE")
```
Kawron: Why do you need `\s*`? Are there logs with different whitespaces than just space, or are there logs that have zero or more than one whitespace char betweeen bracket and the field name? If not then replace it with just space: `*\{ IP \}.*`

Kawron (changes requested):

Kawron (changes requested):
https://github.com/splunk/splunk-connect-for-syslog/pull/3067#issuecomment-4450384832

Kawron (changes requested):
Regex is different between main and lite versions

Kawron (approved):

Kawron (approved):

Issue #2011, opened 2023-03-01
Helm Chart Service externalTrafficPolicy


roosterfish:
Hi,

When trying to use the Helm chart with a Service of type NodePort, it would be beneficial being able to set the `externalTrafficPolicy` setting through the `values.yml` file. Currently this is hardcoded to `Local`. Setting it to `Cluster` allows to make use of the traffic proxying feature between compute nodes of the cluster.

* https://github.com/splunk/splunk-connect-for-syslog/blob/main/charts/splunk-connect-for-syslog/templates/service.yaml#L17
* https://github.com/splunk/splunk-connect-for-syslog/blob/main/charts/splunk-connect-for-syslog/templates/service.yaml#L95

Is there any reason this is currently hardcoded?

Thanks!

rjha-splunk:
Hi,

Thanks for raising it , we will take this and review in our upcoming sprint.

The values.yaml is still evolving , when we see a chance we improve it.

Regards,
Rahul

ehlo550:
Hi, 
From my point of view this is hardcoded to local as you would loose the original source IP of the logs when using Cluster.

roosterfish:
Hi @ehlo550, you will indeed loose the IP information, if the traffic doesn't hit a pod on the same compute node you are connecting to. However having this configurable wouldn't require putting an additional LB in front (e.g. if MetalLB is not an option) since the cluster will take care about distributing the incoming traffic to active pods all over the cluster.

sbylica-splunk:
This will be a part of our next release

sbylica-splunk:
Added in v3.44.0 https://github.com/splunk/splunk-connect-for-syslog/releases/tag/v3.44.0

PR #3078, opened 2026-05-26
fix: Helm Chart Service externalTrafficPolicy

sbylica-splunk:
Implements https://github.com/splunk/splunk-connect-for-syslog/issues/2011

### DIFF
--- a/charts/splunk-connect-for-syslog/values.yaml
+++ b/charts/splunk-connect-for-syslog/values.yaml
@@ -191,6 +191,13 @@ service:
   type: LoadBalancer
   usemetallb: {}
   metallbsharingkey: {}
+  # externalTrafficPolicy controls how external traffic is routed to pods on
+  # Service types `LoadBalancer` and `NodePort`. Defaults to `Local` to
+  # preserve the original client source IP, which is required for correct
+  # `HOST` enrichment of incoming syslog messages. Set to `Cluster` to allow
+  # kube-proxy to load-balance traffic across all nodes at the cost of losing
+  # the original source IP (kube-proxy will SNAT the traffic).
+  externalTrafficPolicy: Local
 
 resources:
   {}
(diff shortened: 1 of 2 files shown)


charts/splunk-connect-for-syslog/values.yaml
```diff
@@ -191,6 +191,13 @@ service:
   type: LoadBalancer
   usemetallb: {}
   metallbsharingkey: {}
+  # externalTrafficPolicy controls how external traffic is routed to pods on
+  # Service types `LoadBalancer` and `NodePort`. Defaults to `Local` to
+  # preserve the original client source IP, which is required for correct
+  # `HOST` enrichment of incoming syslog messages. Set to `Cluster` to allow
+  # kube-proxy to load-balance traffic across all nodes at the cost of losing
```
Kawron, line 198: I think that techniclly it is possible to retain source IP in Cluster mode. In Cilium you can use DSR mode https://docs.cilium.io/en/latest/network/kubernetes/kubeproxy-free/#client-source-ip-preservation. But I don't know if we want to touch this topic in this PR. It may be to complicated. Note: I didn't test that

Kawron (approved):

charts/splunk-connect-for-syslog/values.yaml
sbylica-splunk: Not sure if this is in scope here, this is based on a "basic" kubernetes docs:
https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip

ajasnosz: If I get it correctly for now, any other option than local would make the source IP swapped in the sc4s logs? Maybe we want to make a note about it in documentation?

sbylica-splunk: Yeah, but 'Local' is the default, same as it was before. We have a note like this: `... at the cost of losing the original source IP...`, do we need to add anything more?

ajasnosz: I was thinking of the docs on gh page, but if we don't really want to expose it and just assume that only the people that want to change it will open the values.yaml we can just leave it here

PR #3088, opened 2026-06-03
fix: Refactor duplicated SC4S parsers

digrajpu:
This PR refactors the SC4S parser packaging structure to remove duplicated parser definitions between the etc and lite packages and establish a single source of truth for parser maintenance.

**Changes**

- Introduced a shared parser structure used by both etc and lite packages.
- Moved vendor-specific parser definitions into the shared package.
- Updated both Docker file to consume parser files from the shared location.
- Preserved the existing runtime directory structure expected by SC4S to avoid functional changes.
- Removed duplicated parser definitions that were previously maintained separately in etc and lite.

**Benefits**

- Eliminates duplicate parser maintenance.
- Reduces the risk of inconsistencies between etc and lite implementations.
- Simplifies future parser development and updates.
- Establishes a single source of truth for parser definitions.

Notes
This change is intended to be a code-organization refactor only and does not introduce any functional changes to parser behavior.


sbylica-splunk:
So we currently have a problem because etc and lite are desynchronised slightly (sometimes people make changes to etc and forget the same file in lite), which is actually one of the improvements we want to make with this change.
Before we refactor we should make sure that both directories are synced.
Currently there are 7 files that differ I think: (app-cisco-cisco_asa.conf, app-cisco-cisco_ftd.conf, app-syslog-netapp_ontap.conf, app-syslog-riverbed_syslog.conf, app-cef-netscout_arbor-edge.conf, app-syslog-pure_storage.conf, app-netsource-aruba_silverpeak.conf). 
So probably the best way of doing this would be to create a separate pr to sync the directories first (so we can make sure that the newest version will make it into the combined repo), and then do this refactor.

sbylica-splunk:
There are some places in docs that reference old architecture (like docs/creating_parsers/index.md and some others).

sbylica-splunk:
Also parsers are included into syslog alphabetically (unfortunately), and this changes the directory structure, so we have to make **really** sure that it still works the same way as before.

digrajpu:
> So we currently have a problem because etc and lite are desynchronised slightly (sometimes people make changes to etc and forget the same file in lite), which is actually one of the improvements we want to make with this change. Before we refactor we should make sure that both directories are synced. Currently there are 7 files that differ I think: (app-cisco-cisco_asa.conf, app-cisco-cisco_ftd.conf, app-syslog-netapp_ontap.conf, app-syslog-riverbed_syslog.conf, app-cef-netscout_arbor-edge.conf, app-syslog-pure_storage.conf, app-netsource-aruba_silverpeak.conf). So probably the best way of doing this would be to create a separate pr to sync the directories first (so we can make sure that the newest version will make it into the combined repo), and then do this refactor.

here is the separate [PR](https://github.com/splunk/splunk-connect-for-syslog/pull/3097/changes) which will sync the directories.

digrajpu:
> There are some places in docs that reference old architecture (like docs/creating_parsers/index.md and some others).

This is done

digrajpu:
> Also parsers are included into syslog alphabetically (unfortunately), and this changes the directory structure, so we have to make **really** sure that it still works the same way as before.

I understand the concern about parser load order changing due to the directory structure update. I have been manually testing the refactored implementation, but I wanted to check if there is a recommended way to validate that the behavior remains identical to the current implementation.

Validating parser routing against a representative set of sample logs be sufficient, or is there another approach you would recommend?

Kawron:
> > Also parsers are included into syslog alphabetically (unfortunately), and this changes the directory structure, so we have to make **really** sure that it still works the same way as before.
> 
> I understand the concern about parser load order changing due to the directory structure update. I have been manually testing the refactored implementation, but I wanted to check if there is a recommended way to validate that the behavior remains identical to the current implementation.
> 
> Validating parser routing against a representative set of sample logs be sufficient, or is there another approach you would recommend?

@sbylica-splunk @digrajpu I believe that if the unit tests passes we should be safe

Kawron (changes requested):
Why there are changes unrelated to this task in this PR e.g. docs/gettingstarted/eks.md? There also merge conflicts. Clearly the git history of this branch is messed up. Fix it before puting this PR to the review.

Also, I think that this branch shouldn't be under fix: tag. But I think it can stay this way.

Regarding syncing parsers across lite and main:
1. is it necessary to keep app-syslog-vmware_cb-protect_example_basic.conf and app-syslog-vmware_cb-protect_example.conf
2. what about: app-syslog-riverbed_syslog.conf I can see that there is parser differentiated betwee two versions, but wasn't addressed in your commit. Was it missed, or is there different reason.

When you will do further changes to this branch, please try to keep git history as clean as you can.

digrajpu:
> Why there are changes unrelated to this task in this PR e.g. docs/gettingstarted/eks.md? There also merge conflicts. Clearly the git history of this branch is messed up. Fix it before puting this PR to the review.
> 
> Also, I think that this branch shouldn't be under fix: tag. But I think it can stay this way.
> 
> Regarding syncing parsers across lite and main:
> 
> 1. is it necessary to keep app-syslog-vmware_cb-protect_example_basic.conf and app-syslog-vmware_cb-protect_example.conf
> 2. what about: app-syslog-riverbed_syslog.conf I can see that there is parser differentiated betwee two versions, but wasn't addressed in your commit. Was it missed, or is there different reason.
> 
> When you will do further changes to this branch, please try to keep git history as clean as you can.

I'll clean up the branch, remove the unrelated changes, resolve the conflicts and will create new PR by creating new branch so that it only contains the parser refactor changes and will close this PR

1. Regarding `app-syslog-vmware_cb-protect_example_basic.conf` and `app-syslog-vmware_cb-protect_example.conf`, both files are located under the documentation/examples area and are not included in the Docker build or loaded by SC4S at runtime. I left them unchanged during the parser synchronization work since they are not part of the active parser set. If needed, I can review whether both example files are still useful and clean them up in a separate change.

2. For `app-syslog-riverbed_syslog.conf`, I intentionally kept the version from lite because it contains more recent changes than the version in main. During the synchronization, I chose the latest implementation to avoid losing those updates

here is new PR link: https://github.com/splunk/splunk-connect-for-syslog/pull/3109
CC: @Kawron @sbylica-splunk

PR #3092, opened 2026-06-08
fix: ILO 5 metadata issue

digrajpu:


Kawron (changes requested):
Can you provide little more context. What has changed, where did you found those new logs. Are the previous one wrong? Or is this update for new version of ILO?

tests/test_hpe_ilo.py
```diff
@@ -26,7 +26,7 @@ def test_hpe_ilo_4(
     epoch = epoch[:-3]
 
     mt = env.from_string(
-        "{{ mark }} {{ iso }}Z {{ host }} #ILO4 - - - Browser login: Administrator - 10.0.0.0(host.domain.local)\n"
+        "{{ mark }} {{ iso }}Z {{ host }} DriveArray: Smart Array - Drive is failed...\n"
```
sbylica-splunk, line 64: Yeah I would also need more contest, but also, what happened to the old test here? Are we not supporting this other log format anymore?

digrajpu:
> Can you provide little more context. What has changed, where did you found those new logs. Are the previous one wrong? Or is this update for new version of ILO?

@Kawron @sbylica-splunk here we have official doc for remote syslog
[https://support.hpe.com/hpesc/public/docDisplay?docId=a00105236en_us&docLocale=en_US&page=GUID-D7147C7F-2016-0901-06D0-00000000129A.html&utm_source=chatgpt.com](https://support.hpe.com/hpesc/public/docDisplay?docId=a00105236en_us&docLocale=en_US&page=GUID-D7147C7F-2016-0901-06D0-00000000129A.html&utm_source=chatgpt.com)

tests/test_hpe_ilo.py
digrajpu: The old test was removed while I was investigating the iLO 5 format from the documentation. Looking at it again, the existing tests cover a different format where the PROGRAM field is #ILO4, which will not match with updated parser. I'll restore the old test and update the parser to maintain backward compatibility.

package/etc/conf.d/conflib/syslog/app-syslog-hpe_ilo.conf
```diff
@@ -14,7 +14,12 @@ block parser app-syslog-hpe_ilo() {
 };
 application app-syslog-hpe_ilo[sc4s-syslog-pgm] {
         filter {
-        program('(#ILO|[iI][lL][oO])\d{1}+')
+            program('(#ILO|[iI][lL][oO])\d+')
+            or
+            match(
+            '^(DriveArray|PowerSupply|Fan|Memory|Processor)'
+            value("PROGRAM")
+        )
```
Kawron: that's a weird formatting, I think it would better if closing parentheses would be on the same col as `m` and the match body to be indented one more tab.

Kawron (approved):

Issue #2648, opened 2024-12-02
Request for Updates to SC4S Build for RKE Cluster Deployment


hatem2018:
Description: I am currently deploying Splunk Connect for Syslog (SC4S) on an RKE cluster and have made some progress by updating the Dockerfile and entrypoint.sh to change the user to syslog. Below are the details of my setup and the steps I followed:

Configuration Details:

Provider: RKE1
Kubernetes Version: v1.28.7
Architecture: Amd64
SC4S Version: 4.8.1
Deployment Method: AzureDevOps pipelines with Helm chart
Private Registry: JFrog

Steps to Reproduce:
[image-sc4s.zip](https://github.com/user-attachments/files/17971561/image-sc4s.zip)
[pipeline-sc4s.zip](https://github.com/user-attachments/files/17971562/pipeline-sc4s.zip)

- Set up an RKE cluster with the above configuration.
- Update the Dockerfile and entrypoint.sh to change the user to syslog.
- Push the custom SC4S image to the JFrog registry using Azure pipelines.
- Modify the image in the values.yaml file to use a custom SC4S image.
- Use AzureDevOps pipelines to deploy SC4S using the Helm chart.

Expected Behavior:  I should be able to change any configuration in the values.yaml file using the syslog user without encountering issues related to creating folders and files inside the pods.

Actual Behavior: For example, when configuring custom index using values.yaml file, the configuration is applied but the pods show error creating folder (see screenshot attached)

Request: I would like to request updates to the SC4S build to better support this deployment strategy. Specifically:

Improvements or changes to streamline the process of changing the user to syslog within the SC4S deployment.
Guidance on using a non-root user, such as the syslog user (UID 1024), as our internal policy does not permit the use of root on cluster pods.

Additional Information:

Attached are the updated Dockerfile, entrypoint.sh file, pipeline, and values.yaml file.
This request is related to the Splunk ODS case opened Number 3623626.

Thank you

hatem2018:
![issue-on-pods](https://github.com/user-attachments/assets/ff2235f1-79e4-4c4e-be88-e0f374fb1da5)

rjha-splunk:
We will check and get back, please expect a reply on it in a week.

sbylica-splunk:
I've added a task to our backlog:
https://splunk.atlassian.net/browse/ADDON-77572

sbylica-splunk:
@hatem2018 do you have a support case for this? Can you share it?

[...]

PR #3103, opened 2026-06-11
feat: add non root support for k8s

sbylica-splunk:
Addresses RKE cluster deployment from this issue:
https://github.com/splunk/splunk-connect-for-syslog/issues/2648

### DIFF
--- a/charts/splunk-connect-for-syslog/templates/_helpers.tpl
+++ b/charts/splunk-connect-for-syslog/templates/_helpers.tpl
@@ -62,22 +62,3 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
-
-{{/*
-Pod-level securityContext. Defaults to an empty context (root, default
-behavior); set .Values.podSecurityContext to run as non-root. See values.yaml
-for the recommended non-root preset.
-*/}}
-{{- define "splunk-connect-for-syslog.podSecurityContext" -}}
-{{- toYaml .Values.podSecurityContext -}}
-{{- end -}}
-
-{{/*
-Container-level securityContext. Defaults to an empty context (root, default
-behavior); set .Values.securityContext to run as non-root. See values.yaml for
-the recommended non-root preset (NET_BIND_SERVICE is needed so the default
-privileged ports 514/601 can bind as a non-root user).
-*/}}
-{{- define "splunk-connect-for-syslog.securityContext" -}}
-{{- toYaml .Values.securityContext -}}
-{{- end -}}
--- a/charts/splunk-connect-for-syslog/templates/_helpers.tpl
+++ b/charts/splunk-connect-for-syslog/templates/_helpers.tpl
@@ -62,3 +62,46 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Pod-level securityContext.
+Precedence: an explicit non-empty .Values.podSecurityContext always wins; else
+when .Values.runAsNonRoot is true a non-root preset is applied; otherwise the
+historical default (root, empty context) is used.
+*/}}
+{{- define "splunk-connect-for-syslog.podSecurityContext" -}}
+{{- if .Values.podSecurityContext -}}
+{{- toYaml .Values.podSecurityContext -}}
+{{- else if .Values.runAsNonRoot -}}
+runAsNonRoot: true
+runAsUser: 1024
+runAsGroup: 1024
+fsGroup: 1024
+seccompProfile:
+  type: RuntimeDefault
+{{- else -}}
+{}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Container-level securityContext.
+Precedence: an explicit non-empty .Values.securityContext always wins; else when
+.Values.runAsNonRoot is true a non-root preset is applied; otherwise the
+historical default (root, empty context) is used. NET_BIND_SERVICE is added so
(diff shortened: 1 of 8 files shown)


charts/splunk-connect-for-syslog/templates/_helpers.tpl
```diff
@@ -62,3 +62,22 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Pod-level securityContext. Defaults to an empty context (root, historical
+behavior); set .Values.podSecurityContext to run as non-root. See values.yaml
+for the recommended non-root preset.
+*/}}
+{{- define "splunk-connect-for-syslog.podSecurityContext" -}}
```
Kawron: I wonder if defining this helper is necessary, as it only wraps .Values.podSecruityContext

charts/splunk-connect-for-syslog/templates/statefulset.yaml
```diff
@@ -37,11 +37,11 @@ spec:
       {{- end }}
       serviceAccountName: {{ include "splunk-connect-for-syslog.serviceAccountName" . }}
       securityContext:
-        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+        {{- include "splunk-connect-for-syslog.podSecurityContext" . | nindent 8 }}
```
Kawron: Again, as in https://github.com/splunk/splunk-connect-for-syslog/pull/3103/changes#r3411615226 what's the benefit of moving it the helper function. Why not add toYaml here?

charts/splunk-connect-for-syslog/values.yaml
```diff
@@ -174,18 +174,33 @@ persistence:
 
 podAnnotations: {}
 
+# securityContext settings. Both default to an empty context, which preserves the
+# default behavior of running SC4S as root (recommended for existing deployments
+# and standard runtimes).
+#
+# To run SC4S as the unprivileged "syslog" user (UID/GID 1024) baked into the
+# image -- required for hardened environments such as RKE2/OpenShift or any
```
Kawron: I would change -- to just -. I think that -- is sometimes used to represent em-dash but I think that in the code we can just use hyphen

charts/splunk-connect-for-syslog/values.yaml
```diff
@@ -174,18 +174,33 @@ persistence:
...
+# cluster that enforces the "restricted" Pod Security Standard -- uncomment the
+# preset below. NET_BIND_SERVICE lets the default privileged ports (514/601) bind
+# as non-root and is the only capability the "restricted" profile permits adding;
```
Kawron: We mention NET_BIND_SERVICE here but we are adding it to secruityContext. I would move this comment down

charts/splunk-connect-for-syslog/values.yaml
```diff
@@ -174,18 +174,33 @@ persistence:
...
+# fsGroup makes the persistent volume writable by the "syslog" group.
 podSecurityContext:
   {}
-  # fsGroup: 2000
+  # runAsNonRoot: true
```
Kawron, line 206: I'm curious what will happen when I won't set `runAsNonRoot` but I will set `runAsUser`?

Kawron (changes requested):

charts/splunk-connect-for-syslog/templates/_helpers.tpl
sbylica-splunk: Yeah, makes sense, I'll drop it

charts/splunk-connect-for-syslog/templates/statefulset.yaml
sbylica-splunk: Same

charts/splunk-connect-for-syslog/values.yaml
sbylica-splunk: Yup, changed it to something more human :D

charts/splunk-connect-for-syslog/values.yaml
sbylica-splunk: Yeah, changed it anyway so now it's not there at all

charts/splunk-connect-for-syslog/values.yaml
sbylica-splunk: It's fine actually, just tested it.
There is one combination that doesn't work: runAsNonRoot: false and runAsUser: 0 . Which makes sense.

charts/splunk-connect-for-syslog/templates/statefulset.yaml
```diff
@@ -52,19 +56,30 @@ spec:
...
+            value: "{{ $lp.tcp | default 514 }}"
+          - name: SC4S_LISTEN_DEFAULT_UDP_PORT
+            value: "{{ $lp.udp | default 514 }}"
+          - name: SC4S_LISTEN_DEFAULT_RFC6587_PORT
+            value: "{{ $lp.rfc6587 | default 601 }}"
+          - name: SC4S_LISTEN_DEFAULT_RFC5426_PORT
+            value: "{{ $lp.rfc5426 | default 601 }}"
           # Inject additional enviroment variable from values.yaml to add proxy support
-          {{- if .Values.sc4s.proxy.http }}
+          {{- with .Values.sc4s.proxy }}
```
sbylica-splunk, line 74: This is a kind of side-fix, since by default this doesn't exist at all and chart fails to install, so you "have to" create an empty proxy dict.

Kawron (approved):

PR #3104, opened 2026-06-12
feat: add sc4s mcp server to the build process

Kawron:
Adding SC4S mcp server to the CI/CD.

Because it's not being merged to the develop branch, the CI/CD pipeline won't trigger. However, I tested it separately and all steps passes.

### DIFF
--- a/.github/workflows/ci-mcp.yaml
+++ b/.github/workflows/ci-mcp.yaml
@@ -150,7 +150,7 @@ jobs:
     needs:
       - meta
     container:
-      image: python:3.14-slim
+      image: python:3.14-bookworm
     steps:
       - name: Checkout
         uses: actions/checkout@v6
--- a/.github/workflows/ci-mcp.yaml
+++ b/.github/workflows/ci-mcp.yaml
@@ -1,0 +1,230 @@
+#   ########################################################################
+#   Copyright 2021 Splunk Inc.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#   ########################################################################
+
+name: ci-mcp
+on:
+  push:
+    branches:
+      - "main"
+      - "releases/*"
+      - "develop"
+      - "next*"
+  pull_request:
+    branches:
+      - "main"
+      - "releases/*"
+      - "develop"
+      - "next*"
+
+permissions:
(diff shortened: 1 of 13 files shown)


.github/workflows/ci-mcp.yaml
```diff
@@ -0,0 +1,230 @@
...
+          severity: 'CRITICAL,HIGH,MEDIUM,LOW'
+          trivyignores: '.trivyignore'
+          scanners: "vuln"
+
+  test-mcp-unit-tests:
+    runs-on: ubuntu-latest
+    needs:
+      - meta
+    container:
+      image: python:3.14-slim
```
sbylica-splunk: What's the upside of this over say python:3.14-bookworm anyway?

docs/mcp_server/installation.md
```diff
@@ -40,27 +72,10 @@ The MCP server is configured through environment variables.
 | `SC4S_MCP_TLS_KEY`          |                         | Path inside the container to the matching PEM-encoded private key.                                                                                                                                                                                                                                                           |
 | `SC4S_MCP_TLS_KEY_PASSWORD` |                         | Optional passphrase for an encrypted private key.                                                                                                                                                                                                                                                                            |
 
-## Build the image
```
sbylica-splunk, line 43: Do we not need this anymore?

sc4s_mcp/pyproject.toml
```diff
@@ -1,11 +1,11 @@
 [tool.poetry]
 name = "sc4s-mcp"
-version = "0.1.0"
+version = "3.44.0"
```
sbylica-splunk, line 3: Do we need this or can we link the version to the main sc4s version?

pyproject.toml
```diff
@@ -22,6 +22,9 @@ cryptography = "^48.0.0"
 flask-wtf = "^1.2.2"
 fastmcp = ">=3.2.0"
 httpx = "^0.28.1"
+authlib = ">=1.7.2"
```
sbylica-splunk, line 25: Check if this is necessary

sbylica-splunk (approved):

pyproject.toml
Kawron: Okay I've checked trhat, it transative dependency of fastMCP. Here it is forced to be >= 1.7.2 because fastmcp requries >=1.6.5

sc4s_mcp/pyproject.toml
Kawron: .releaserc updates both SC4S pyproject.toml and sc4s_mcp/pyproject.toml. So those two versions will always be in the sync. I don't know if it is possible in .toml to set version to a version from different project. So the current approach is ok

docs/mcp_server/installation.md
Kawron: Now the image is being built via ci/cd, no need to document manual build process (we don't do it for main sc4s as well). Beside it's not that complicated so if someone want to build custom version they should be able to figure it out

.github/workflows/ci-mcp.yaml
Kawron: less packages. But I've changed that to bookwarm for consistency

PR #3119, opened 2026-06-25
fix: some fixes for sonarqube issues

sbylica-splunk:
Scan:
https://sonarqube.splunkdev.net/dashboard?id=github-mirrors.splunk-connect-for-syslog&branch=fix%2Ffix_sonarqube_issues

### DIFF
--- a/.github/workflows/agreements.yaml
+++ b/.github/workflows/agreements.yaml
@@ -5,7 +5,8 @@ on:
   pull_request_target:
     types: [opened, closed, synchronize]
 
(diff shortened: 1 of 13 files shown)


sbylica-splunk:
Mostly smal changes, two files reworked the most are:
* configuration tool - checked it manually and it works without any issues
* pcap sender - also checked it, with both udp and tcp, sends data without issue as far as I can tell

Kawron (approved):

PR #3132, opened 2026-07-03
fix: Fix baremetal release

digrajpu:


.github/workflows/cd-baremtal.yaml
```diff
@@ -47,6 +47,12 @@ jobs:
           tar rvf /tmp/baremetal.tar -C . pyproject.toml
           tar rvf /tmp/baremetal.tar -C . poetry.lock
           tar rvf /tmp/baremetal.tar -C package/sbin .
+          tar rvf /tmp/baremetal.tar --transform 's,^\.,conf.d,' -C package/shared/conf.d .
```
Kawron, line 50: Why do we need to use transofrm here?

Kawron (changes requested):

.github/workflows/cd-baremtal.yaml
digrajpu: We use --transform to prepend the destination directory prefix to each archived file path. Since -C <dir> . stores entries with a leading . the transform rewrites it to the correct path (conf.d/ or conf.d/conflib/) so files extract to the right location.

sbylica-splunk:
I tested it manually and it seems to work, so I'm leaving an approve

sbylica-splunk (approved):

.github/workflows/cd-baremtal.yaml
sbylica-splunk: Yeah, not sure if this is the prettiest way of doing it, but it seems to be working

Kawron (approved):

PR #3136, opened 2026-07-07
fix: Add tests for baremetal package

digrajpu:
- The baremetal release was broken after code refactor— shared/conf.d and shared/addons contents were missing from the tar.
- Added tests/test_baremetal_package.py to prevent regressions — verifies tar size, required files/dirs, syslog-ng @include coverage, addon parity, and that conf.d content does not leak to the archive root.
- Run pytest tests/test_baremetal_package.py  on Ubuntu
- Verify all 10 test groups pass

### DIFF
--- a/.github/workflows/cd-baremtal.yaml
+++ b/.github/workflows/cd-baremtal.yaml
@@ -56,8 +56,8 @@ jobs:
           sha512sum -b /tmp/baremetal.tar > /tmp/baremetal_checksum.txt
       - name: Validate BYOE package
         run: |
-          pip3 install pytest
-          BAREMETAL_TAR=/tmp/baremetal.tar pytest tests/test_baremetal_package.py -v
+          poetry install
+          BAREMETAL_TAR=/tmp/baremetal.tar poetry run pytest tests/test_baremetal_package.py -v
       - name: Release
         uses: softprops/action-gh-release@v3
         with:
--- a/.github/workflows/cd-baremtal.yaml
+++ b/.github/workflows/cd-baremtal.yaml
@@ -54,6 +54,10 @@ jobs:
           tar rvf /tmp/baremetal.tar -C package/shared local_config
           tar rvf /tmp/baremetal.tar -C package/shared test_parsers
           sha512sum -b /tmp/baremetal.tar > /tmp/baremetal_checksum.txt
+      - name: Validate BYOE package
+        run: |
+          pip3 install pytest
+          BAREMETAL_TAR=/tmp/baremetal.tar pytest tests/test_baremetal_package.py -v
       - name: Release
         uses: softprops/action-gh-release@v3
         with:
--- a/.github/workflows/cd-baremtal.yaml
+++ b/.github/workflows/cd-baremtal.yaml
@@ -57,7 +57,7 @@ jobs:
       - name: Validate BYOE package
         run: |
           poetry install
-          BAREMETAL_TAR=/tmp/baremetal.tar poetry run pytest tests/test_baremetal_package.py -v
+          BAREMETAL_TAR=/tmp/baremetal.tar poetry run pytest tests/test_baremetal_package.py -v -m baremetal
       - name: Release
         uses: softprops/action-gh-release@v3
         with:
--- a/tests/test_baremetal_package.py
+++ b/tests/test_baremetal_package.py
@@ -1,4 +1,4 @@
-# Copyright 2019 Splunk, Inc.
+# Copyright 2026 Splunk, Inc.
 #
 # Use of this source code is governed by a BSD-2-clause-style
 # license that can be found in the LICENSE-BSD2 file or at
@@ -8,7 +8,7 @@
 Unit tests for the baremetal tar package contents.
 
 These tests verify that the baremetal.tar produced by cd-baremtal.yaml contains
-all required paths and mirrors the layout used by the OCI container (package/Dockerfile).
+all required paths used by the OCI container (package/Dockerfile).
 
 No Docker, Splunk, or network access required — runs anywhere with Python + GNU tar.
 """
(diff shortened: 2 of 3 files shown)


tests/test_baremetal_package.py
```diff
@@ -0,0 +1,384 @@
...
+    "enrich",
+    "log_paths",
+    "sources",
+    "sc4slib",
+    "plugin",
+    "conflib",
+]
+
+
+def build_baremetal_tar(dest: str) -> None:
```
sbylica-splunk: I don't think this is a good way of testing it, we are just copying commands from cd-baremtal.yaml, but what if the commands change at some point?

We should probably test the effects of cd-baremtal.yaml itself.

tests/test_baremetal_package.py
```diff
@@ -0,0 +1,384 @@
...
+        os.path.relpath(os.path.join(r, f), src_dir)
+        for r, _, files in os.walk(src_dir)
+        for f in files
+        if not f.endswith(".pyc") and "__pycache__" not in r
+    }
+
+
+# Fixtures — build once per test session
+@pytest.fixture(scope="module")
+def baremetal_tar(tmp_path_factory):
```
sbylica-splunk: We could add a step like:
      - name: Validate BYOE package   
To cd-baremetal, and test a real package here.

sbylica-splunk (changes requested):

tests/test_baremetal_package.py
```diff
@@ -0,0 +1,388 @@
...
+# Directories that must exist at root level.
+REQUIRED_TOP_LEVEL_DIRS = [
+    "pylib",
+    "context_templates",
+    "local_config",
+    "test_parsers",
+]
+
+# These dirs must NOT appear at the top level — they belong under conf.d/.
+# Their presence at root was the exact symptom of the v3.45.0 regression.
```
Kawron: That is unnecessary info

tests/test_baremetal_package.py
```diff
@@ -0,0 +1,388 @@
...
+        ["tar", "rvf", dest, "-C", "package/shared", "pylib"],
+        ["tar", "rvf", dest, "-C", "package/shared", "context_templates"],
+        ["tar", "rvf", dest, "-C", "package/shared", "local_config"],
+        ["tar", "rvf", dest, "-C", "package/shared", "test_parsers"],
+    ]
+    for cmd in cmds:
+        subprocess.run(cmd, cwd=REPO_ROOT, check=True, capture_output=True)
+
+
+def is_gnu_tar() -> bool:
```
Kawron: I don't think it is necessary

tests/test_baremetal_package.py
```diff
@@ -0,0 +1,388 @@
...
+def _collect_source_files(src_dir: str) -> set:
+    return {
+        os.path.relpath(os.path.join(r, f), src_dir)
+        for r, _, files in os.walk(src_dir)
+        for f in files
+        if not f.endswith(".pyc") and "__pycache__" not in r
+    }
+
+
+# Fixtures — build once per test session
```
Kawron: I'm personally not a big fun of such comments. It is clearly AI trying to explain its own code and according to the self-explaining code rule this is not necessary. I would assume that developer (or AI) working on this project will have a basic knowledge of pytest fixtures

tests/test_baremetal_package.py
```diff
@@ -0,0 +1,388 @@
...
+@pytest.fixture(scope="module")
+def baremetal_tar(tmp_path_factory):
+    if not is_gnu_tar():
+        pytest.skip("GNU tar required (CI runs Ubuntu; use Docker locally)")
```
Kawron: Why skip? In this case I would fail. It's not very usefull otherwise

tests/test_baremetal_package.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+# 1. Size sanity test — catches a completely empty package
+def test_tar_size_indicates_parsers_are_present(baremetal_tar):
+    """
+    Because all parsers were missing.
+    A correct package must be larger than 500KB.
+    """
+    size = os.path.getsize(baremetal_tar)
+    assert size > 500_000, (
```
Kawron: Hard coded value. Seems arbitrary. Why check it if we are later testing if all the necessary files exists

tests/test_baremetal_package.py
```diff
@@ -0,0 +1,388 @@
...
+        if not conf_files:
+            empty.append(required_dir)
+    assert not empty, (
+        f"These @include directories exist but contain no .conf files: {empty}. "
+        "syslog-ng will start but silently have no parsers or routing."
+    )
+
+
+# 4. Addon test — every addon in shared/addons must land in conf.d/conflib
+def test_every_addon_dir_present_in_conflib(extracted_tar):
```
Kawron, line 131: here you could check if those file are not only missing but also if they are not empty

.github/workflows/cd-baremtal.yaml
```diff
@@ -54,6 +54,10 @@ jobs:
           tar rvf /tmp/baremetal.tar -C package/shared local_config
           tar rvf /tmp/baremetal.tar -C package/shared test_parsers
           sha512sum -b /tmp/baremetal.tar > /tmp/baremetal_checksum.txt
+      - name: Validate BYOE package
```
Kawron, line 57: Have you tested if this will run? I think it will fail as you don't install all dependencies.

See other workflows, they use poetry install

tests/test_baremetal_package.py
Kawron: Exactly we are duplicating logic here. And adding a second source of truth. It is very dangerous as in the future if sth breaks we won't know what is the correct implementation. I don't think this issue was addressed

Kawron (changes requested):
I don't think this is a right approach. Why just don't build the sc4s baremetal and verify it's working?

Kawron (approved):

sbylica-splunk (approved):

PR #3143, opened 2026-07-13
Feat/mcp multi client support

Kawron:
Issue: Two MCP clients could modify the SC4S configuration simultaneously, potentially causing requests to be queued and updates to be lost.

Fix: Add checks to ensure that Gunicorn runs with a single thread. Convert endpoints that require a syslog-ng restart to asynchronous operations. If another job is already running, return HTTP 409.

### DIFF
--- a/package/sbin/config_api.py
+++ b/package/sbin/config_api.py
@@ -4,6 +4,7 @@ from flask_wtf.csrf import CSRFProtect
 from flask import Blueprint, jsonify, request
 
 from constants import ENV_FILE, PARSERS_DIR
+from job_api import submit_job
 from utils import apply_with_rollback
 
 logger = logging.getLogger(__name__)
@@ -35,12 +36,13 @@ def set_env():
     if not file.filename:
         return jsonify({"status": "error", "message": "empty file"}), 400
 
-    try:
-        apply_with_rollback({ENV_FILE: file.read().decode("utf-8")})
-    except Exception as e:
-        return jsonify({"status": "error", "message": str(e)}), 500
+    content = file.read().decode("utf-8")
 
-    return jsonify({"status": "env_file updated successfully"}), 200
+    def apply_env():
+        apply_with_rollback({ENV_FILE: content})
+        return {"status": "env_file updated successfully"}
+
+    return submit_job(apply_env)
 
 
 @csrf.exempt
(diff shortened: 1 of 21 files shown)


package/sbin/job_api.py
```diff
@@ -0,0 +1,73 @@
...
+                    "message": str(exc),
+                    "job_id": exc.job_id,
+                    "url": _get_status_url(exc.job_id),
+                }
+            ),
+            500,
+        )
+
+    url = _get_status_url(job.job_id)
+    response = jsonify({"status": "success", "job_id": job.job_id})
```
digrajpu: A 202 means the job is queued, not finished yet, so returning 'status': 'success' might confusing. I think It should be changed to 'accepted' or 'in_progress'.

package/sbin/utils.py
```diff
@@ -37,6 +73,23 @@ def restart_syslog_ng():
     if result.returncode != 0:
         raise RuntimeError(f"syslog-ng restart failed: {result.stderr.strip()}")
 
+    deadline = time.time() + RESTART_TIMEOUT_SECONDS
+    while time.time() < deadline:
+        current_pids = _get_syslog_ng_pids()
+        if current_pids - previous_pids and _syslog_ng_is_healthy():
+            if time.time() < deadline:
+                return
```
digrajpu: The inner if time.time() < deadline condition is always true because it duplicates the outer while loop condition. The break below it is unreachable dead code. Just return directly when the health check passes instead.

package/sbin/job_api.py
Kawron: Done

package/sbin/utils.py
Kawron: Technically, the deadline timeout could pass till we get to this condition. Though, I simplified it a little bit

digrajpu (approved):

package/sbin/config_api.py
```diff
@@ -54,15 +59,13 @@ def add_parser():
...
 
     parser_path = PARSERS_DIR / file.filename
+    content = file.read().decode("utf-8")
 
-    try:
-        apply_with_rollback({parser_path: file.read().decode("utf-8")})
-    except Exception as e:
-        return jsonify({"status": "error", "message": str(e)}), 500
+    def apply_parser():
+        apply_with_rollback({parser_path: content})
```
sbylica-splunk, line 65: I think we are missing exception handling here now?

sbylica-splunk (commented):
Looks good to me, only one comment

sbylica-splunk (approved):

package/sbin/config_api.py
Kawron: This is an inner function that's later passed to `apply_with_rollback`, later in the `job_manager` we handle the exception and mark it as failed

PR #3150, opened 2026-07-16
fix: Parse high precision timestamp in new Palo Alto Network logs

digrajpu:
PAN (Palo Alto Networks) syslog timestamps don't include timezone information by default, which causes inaccurate event timestamps in Splunk. Starting from newer PAN-OS versions, a high_res_time field was added in each log entry.
PR #2728 introduced parsing of `high_res_time` for 5 log types: THREAT, TRAFFIC, CONFIG, HIPMATCH, USERID. This PR extends that fix to the remaining log types.

FYI: https://splunk.atlassian.net/browse/ADDON-77075

**Changes to app-syslog-pan_panos.conf**

**GLOBALPROTECT** : The existing parser already had a field at position 37 named event_time, which PAN-OS docs label "High Resolution Timestamp". It was never mapped to high_res_time so the shared date-parser at the bottom of the block never picked it up. Converted to if/elif pattern, renaming that field to high_res_time so the timestamp is now correctly parsed.

**AUTH**: Same issue as GLOBALPROTECT. The field at position 34 was named timestamp instead of high_res_time. Converted to if/elif pattern with the correct field name.

**SCTP, GTP, TUNNEL INSPECTION**: These three log types had no parser at all in sc4s. New parser blocks added based on PAN-OS 10.2 field documentation, with high_res_time. Field counts verified against the official docs.

**Log types not requiring parser changes**

- URL, DATA FILTERING Subtypes of THREAT (log_subtype=url/data). Already handled by THREAT parser from PR #2728.
- IPTAG Subtype of USERID (log_subtype=Iptag). Already handled by USERID parser from PR #2728.
- DECRYPT Already updated in PR #2728 as DECRYPTION.
- SYSTEM Already updated in this codebase (SYSTEM parser has if/elif with high_res_time).
- CORRELATED EVENTS PAN-OS docs confirm this log type has no high_res_time field.
- AUDIT Subtype of SYSTEM (log_subtype=AUDIT). Handled by SYSTEM parser. No high_res_time field per PAN-OS docs.

**New tests added (tests/test_palo_alto.py)**

test_palo_alto_globalprotect_high_resolution_timestamp
test_palo_alto_auth_high_resolution_timestamp
test_palo_alto_sctp
test_palo_alto_gtp
test_palo_alto_tunnel

Each test sends a log with a high_res_time offset 15 minutes behind the syslog header timestamp, then verifies the event lands in Splunk at _time matching the high_res_time value. confirming that the high-precision timestamp takes priority over the BSD header time.

### DIFF
--- a/package/shared/addons/palo-alto/app-syslog-pan_panos.conf
+++ b/package/shared/addons/palo-alto/app-syslog-pan_panos.conf
@@ -203,14 +203,26 @@ block parser app-syslog-pan_panos() {
                 );
             };
         } elif (message(',GLOBALPROTECT,' type(string) flags(substring))) {
-            parser {
-                csv-parser(
-                    columns("future_use1","receive_time","serial_number","log_type","future_use2","version","time_generated","vsys","event_id","stage","auth_method","tunnel_type","src_user","src_region","machine_name","public_ip","public_ipv6","private_ip","private_ipv6","host_id","serial_number","client_ver","client_os","client_os_ver","repeat_count","reason","error","opaque","status","location","login_duration","connect_method","error_code","portal","sequence_number","action_flags","event_time","selection_type","response_time","priority","attempted_gateways","gateway","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name","vsys_id")
-                    prefix(".values.")
-                    delimiters(',')
-                    quote-pairs('""')
-                    flags(escape-double-char)
-                );
+            if {
+                parser {
+                    csv-parser(
+                        columns("future_use1","receive_time","serial_number","log_type","future_use2","version","time_generated","vsys","event_id","stage","auth_method","tunnel_type","src_user","src_region","machine_name","public_ip","public_ipv6","private_ip","private_ipv6","host_id","serial_number","client_ver","client_os","client_os_ver","repeat_count","reason","error","opaque","status","location","login_duration","connect_method","error_code","portal","sequence_number","action_flags","high_res_time","selection_type","response_time","priority","attempted_gateways","gateway","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name","vsys_id")
+                        prefix(".values.")
+                        delimiters(',')
+                        quote-pairs('""')
+                        flags(escape-double-char)
+                    );
+                };
+            } elif {
+                parser {
+                    csv-parser(
+                        columns("future_use1","receive_time","serial_number","log_type","future_use2","version","time_generated","vsys","event_id","stage","auth_method","tunnel_type","src_user","src_region","machine_name","public_ip","public_ipv6","private_ip","private_ipv6","host_id","serial_number","client_ver","client_os","client_os_ver","repeat_count","reason","error","opaque","status","location","login_duration","connect_method","error_code","portal","sequence_number","action_flags","event_time","selection_type","response_time","priority","attempted_gateways","gateway","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name","vsys_id")
+                        prefix(".values.")
+                        delimiters(',')
+                        quote-pairs('""')
+                        flags(escape-double-char)
+                    );
+                };
             };
             rewrite {
                 r_set_splunk_dest_update_v2(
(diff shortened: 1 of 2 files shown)


Kawron:
@digrajpu can you provide more detail in the PR description what has changed and what is the reason for such change?

Kawron (changes requested):
What do we want to do about this (excerpt from https://docs.paloaltonetworks.com/ngfw/administration/monitoring/use-syslog-for-monitoring/syslog-field-descriptions/traffic-log-fields):
> The High Resolution Timestamp is supported for logs received from managed firewalls running PAN-OS 10.0 and later releases. Logs received from managed firewalls running PAN-OS 9.1 and earlier releases display a 1969-12-31T16:00:00:000-8:00 timestamp regardless of when the log was received.

FYI: @sbylica-splunk

package/shared/addons/palo-alto/app-syslog-pan_panos.conf
```diff
@@ -203,14 +203,26 @@ block parser app-syslog-pan_panos() {
...
-                    prefix(".values.")
-                    delimiters(',')
-                    quote-pairs('""')
-                    flags(escape-double-char)
-                );
+            if {
+                parser {
+                    csv-parser(
+                        columns("future_use1","receive_time","serial_number","log_type","future_use2","version","time_generated","vsys","event_id","stage","auth_method","tunnel_type","src_user","src_region","machine_name","public_ip","public_ipv6","private_ip","private_ipv6","host_id","serial_number","client_ver","client_os","client_os_ver","repeat_count","reason","error","opaque","status","location","login_duration","connect_method","error_code","portal","sequence_number","action_flags","high_res_time","selection_type","response_time","priority","attempted_gateways","gateway","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name","vsys_id")
```
Kawron, line 209: why `time_generated` when in line 365 we have `generated_time`. It seems that this is an bug. Also we should add tests for testing this if/elif branch making sure that events with missing high_res timestamp will be parsed correctly as well

tests/test_palo_alto.py
```diff
@@ -592,4 +586,148 @@ def test_palo_alto_decryption_high_resolution_timestamp(record_property,  setup_
...
+    result_count, _ = splunk_single(setup_splunk, search)
+
+    record_property("host", host)
+    record_property("resultCount", result_count)
+    record_property("message", message)
+
+    assert result_count == 1
+
+
+# <190>Mar 28 05:40:45 system-host 1,2025/03/28 05:40:45,000000000000,SCTP,0,,2025/03/28 05:40:45,10.0.0.1,10.0.0.2,,,,,,,,vsys1,trust,untrust,ethernet1/1,ethernet1/2,log-action,,1111,1,1111,2905,,,,sctp,allow,0,0,0,0,vsys1,system-host,1111111,,,1111111,,2,sctp-data,sctp-init,tag1,tag2,,,,,,,,,,,,,,,,,,0,0,0,rule-uuid,2025-03-28T05:40:45.986-04:00
```
sbylica-splunk: I think there is something wrong with this example (only the comment, real field below seems to be ok). It has 70 fields and the parser and actual log have 65. Can you re-check this example?

tests/test_palo_alto.py
```diff
@@ -592,4 +586,148 @@ def test_palo_alto_decryption_high_resolution_timestamp(record_property,  setup_
...
+    result_count, _ = splunk_single(setup_splunk, search)
+
+    record_property("host", host)
+    record_property("resultCount", result_count)
+    record_property("message", message)
+
+    assert result_count == 1
+
+
+# <190>Mar 28 05:40:45 system-host 1,2025/03/28 05:40:45,000000000000,GTP,0,,2025/03/28 05:40:45,10.0.0.1,10.0.0.2,,,rule1,,,web-browsing,vsys1,trust,untrust,ethernet1/1,ethernet1/2,log-action,,1111,,1111,2152,,,,,tcp,allow,gtp-create-session,441234567890,internet.apn,lte,create-session-request,192.168.1.1,teid1,teid2,s11,0,medium,310,260,111,22222,1,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,60,3600,tunnel-rule,10.1.1.1,user,rule-uuid,pcap-id,2025-03-28T05:40:45.986-04:00,,,management,general-internet,networking,2,has-known-vulnerability,web-browsing,no,no
```
sbylica-splunk: Same here: 114 fields as opposed to 94 in a real log

package/shared/addons/palo-alto/app-syslog-pan_panos.conf
```diff
@@ -248,9 +260,38 @@ block parser app-syslog-pan_panos() {
...
+                        prefix(".values.")
+                        delimiters(',')
+                        quote-pairs('""')
+                        flags(escape-double-char)
+                    );
+                };
+            } elif {
+                parser {
+                    csv-parser(
+                        columns("future_use1","receive_time","serial_number","log_type","log_subtype","version","generated_time","vsys","src_ip","user","user_normalized","object","authentication_policy","repeast_count","authentication_id","pan_vendor","log_action","server_profile","description","client_type","event_type","factor_number","sequence_number","action_flags","device_group_hierarchy_1","device_group_hierarchy_2","device_group_hierarchy_3","device_group_hierarchy_4","vsys","dvc_name","vsys_id","authentication_protocol","rule","timestamp","src_host_category","src_host_profile","src_host_model","src_host_vendor","src_host_os_name","src_host_os_version","src_host","src_mac","region","future_use2","user_agent","session_id","cluster_name")
```
sbylica-splunk: repeast_count looks like a typo

sbylica-splunk (changes requested):

digrajpu:
> What do we want to do about this (excerpt from https://docs.paloaltonetworks.com/ngfw/administration/monitoring/use-syslog-for-monitoring/syslog-field-descriptions/traffic-log-fields):
> 
> > The High Resolution Timestamp is supported for logs received from managed firewalls running PAN-OS 10.0 and later releases. Logs received from managed firewalls running PAN-OS 9.1 and earlier releases display a 1969-12-31T16:00:00:000-8:00 timestamp regardless of when the log was received.
> 
> FYI: @sbylica-splunk

I handled this by setting `high_res_time` to an empty string. if it contains '1969' in `high_res_time` field. allowing it to fall back to generated_time

package/shared/addons/palo-alto/app-syslog-pan_panos.conf
digrajpu: This is not a bug GLOBALPROTECT uses a time_generated field. and I've also added the test cases that don't have the high_res timestamp field

tests/test_palo_alto.py
digrajpu: done

tests/test_palo_alto.py
digrajpu: Done

package/shared/addons/palo-alto/app-syslog-pan_panos.conf
digrajpu: Done

package/shared/addons/palo-alto/app-syslog-pan_panos.conf
Kawron: This will always suceed (https://github.com/splunk/splunk-connect-for-syslog/blob/develop/package/shared/conf.d/conflib/_common/p_date-parser_nofilter.conf#L1-L21 because of the else) so the elif branch (in this code) won't execute.

The tests on 414 line doesn't find this because they use the same time for bsd and high_res timestamp, change that.


I would use just date-parser, or if necessary createa another wrapper for date-parser that will actually fail

Kawron (changes requested):

package/shared/addons/palo-alto/app-syslog-pan_panos.conf
digrajpu: done

Kawron (approved):

sbylica-splunk (approved):

PR #3151, opened 2026-07-16
feat: added config setting for listen status host

sbylica-splunk:
Addresses the `Avoid binding the application to all network interfaces` security issue: now this is available as a configurable option instead of hardcoded 0.0.0.0 host.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -344,9 +344,11 @@ ExecStart=/usr/bin/podman run \
         --name SC4S \
         --rm $SC4S_IMAGE
 ```
-# Change your status port
+# Change your status host and port
 
 Use `SC4S_LISTEN_STATUS_PORT` to change the "status" port used by the internal health check process. The default value is `8080`.
+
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -344,10 +344,12 @@ ExecStart=/usr/bin/podman run \
         --name SC4S \
         --rm $SC4S_IMAGE
 ```
-# Change your status port
+# Change your status host and port
 
 Use `SC4S_LISTEN_STATUS_PORT` to change the "status" port used by the internal health check process. The default value is `8080`.
 
+Use `SC4S_LISTEN_STATUS_HOST` to change the network interface the internal health check process binds to. The default value is `0.0.0.0`, which binds to all interfaces so the health endpoint is reachable by container orchestrators (for example, Kubernetes liveness/readiness probes and Docker port mappings). If you only need local access, set this to `127.0.0.1` to restrict the endpoint to the local interface.
```
Kawron, line 351: Can we change the naming here to sth like SC4S_LISTEN_HOST and SC4S_LISTEN_PORT.

When we will release MCP it will no longer be only healtcheck, but the whole managment API (of course when enabled)

Kawron (changes requested):

docs/configuration.md
sbylica-splunk: I would avoid it to keep things backwards compatible to be honest. Maybe once we have MCP we can enable both versions?

Kawron (approved):

PR #3166, opened 2026-07-27
Feat/improve memory usage

Kawron:
This PR does the refactor of vps and name caches.
1. Remove cirucal dependencies with the traceback function
2. Implement proper batching for sqlite
3. Implement proper flow control as described in syslog-ng documentation

syslog-ng documentation: https://syslog-ng.github.io/admin-guide/070_Destinations/200_Python/README.html#sendself-message-method-mandatory

### DIFF
--- a/package/shared/pylib/parser_cef.py
+++ b/package/shared/pylib/parser_cef.py
@@ -1,5 +1,4 @@
 import re
-import sys
 import traceback
 
 try:
--- a/package/shared/pylib/parser_cef.py
+++ b/package/shared/pylib/parser_cef.py
@@ -66,9 +66,7 @@ class cef_kv(LogParser):
                 log_message[f".values.{kc}"] = v
 
         except Exception:
-            exc_type, exc_value = sys.exc_info()[:2]
-            lines = traceback.format_exception(exc_type, exc_value)
-            self.logger.debug("".join("!! " + line for line in lines))
+            self.logger.debug(traceback.format_exc())
             return False
 
         return True
--- a/package/shared/pylib/parser_cef.py
+++ b/package/shared/pylib/parser_cef.py
@@ -66,8 +66,8 @@ class cef_kv(LogParser):
                 log_message[f".values.{kc}"] = v
 
         except Exception:
-            exc_type, exc_value, exc_traceback = sys.exc_info()
-            lines = traceback.format_exception(exc_type, exc_value, exc_traceback)
+            exc_type, exc_value = sys.exc_info()[:2]
+            lines = traceback.format_exception(exc_type, exc_value)
             self.logger.debug("".join("!! " + line for line in lines))
             return False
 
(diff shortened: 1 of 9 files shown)


package/shared/pylib/parser_source_cache.py
```diff
@@ -102,24 +105,28 @@ def send(self, log_message):
...
+            except KeyError:
+                self.db[ip_int] = log_message["HOST"]
+            else:
                 if current != log_message["HOST"]:
                     self.db[ip_int] = log_message["HOST"]
-            else:
-                self.db[ip_int] = log_message["HOST"]
-
+        except KeyError:
+            return self.ERROR
```
sbylica-splunk: Do we want ERROR or DROP here? Per doc:
https://syslog-ng.github.io/admin-guide/070_Destinations/200_Python/README.html#sendself-message-method-mandatory

This probably will just get retried and throw exception again?

package/shared/pylib/parser_cef.py
sbylica-splunk: Some unused imports left

package/shared/pylib/parser_source_cache.py
Kawron: I changed it to Success

digrajpu (approved):

package/shared/pylib/parser_source_cache.py
```diff
@@ -81,45 +79,74 @@ def parse(self, log_message):
...
+
+            try:
+                host = log_message["HOST"]
+            except KeyError:
+                self.logger.debug("psc.send skipped: HOST is missing")
+                return self.SUCCESS
+
+            if not host:
+                self.logger.debug(
+                    f"psc.send skipped: HOST is empty"
```
sbylica-splunk: This f string doesn't need to be an f string

package/shared/pylib/parser_source_cache.py
```diff
@@ -81,45 +79,74 @@ def parse(self, log_message):
...
         except Exception:
-            exc_type, exc_value, exc_traceback = sys.exc_info()
-            lines = traceback.format_exception(exc_type, exc_value, exc_traceback)
-            self.logger.debug("".join("!! " + line for line in lines))
-            return False
+            self.logger.debug(traceback.format_exc())
+            return self.ERROR
         self.logger.debug("psc.send complete")
-        return True
+        return self.QUEUED
```
sbylica-splunk, line 141: Ok since we are changing the behaviour a bit/returning new statuses etc, maybe we can add some new tests? I'm approving here because it looks good but we can still add some.

sbylica-splunk (approved):

package/shared/pylib/parser_source_cache.py
sbylica-splunk: Oh one more thing, since we are returning statuses here, can we also rework other pylib parsers to return statuses too? From what I can see they are left with booleans.

Kawron: https://syslog-ng.github.io/admin-guide/120_Parser/018_Python_parser#parseself-log_message - according to the docs parse() should return boolean

Issue #2424, opened 2024-04-25
Configuration of SC4S hosted on ECS/Fargate


krshx:
**What is the sc4s version ?**
Latest version

**Is there a pcap available?**
N/A

**What the vendor name?**
N/A

**What's the product name?**
N/A

** Feature Request description: **
Support to run SC4S on ECS/Fargate deployment/configuration

** Should it support TCP or UDP?**
N/A

** Do you want to have it for local usage or prepare a github PR? **
Github PR will be great as it would be officially supported by splunk, but happy to start with local usage
We have an overarching principle to lean on native aws serverless services and in this case we are trying to leverage aws-ecs & aws-fargate.

Can you kindly confirm/support this configuration mechanism? I.e. deploy/run the sc4s containers (docker based) orchestrated by aws-ecs with fargate-launchtype please? Currently we are facing errors while sc4s is trying to start.

Error Log
* *
2024-04-23T15:46:41.207+02:00 starting syslog-ng
[2024-04-23T13:46:43.079546] Failed to grab disk-buffer dirlock; filename='/var/lib/syslog-ng/syslog-ng-disk-buffer.dirlock', error='Bad file descriptor (9)'
[2024-04-23T13:46:43.079546] Error initializing message pipeline; plugin_name='http', location='root generator dest_hec:5:5'
Handling exit 1 and restarting
 * *

fpunzohig:
Can someone provide an e.t.a. on this feature? We requested it as well in 2022.

rjha-splunk:
Hi @fpunzohig We understand this, we are prioritising this in this quarter and we will update.

jalecrim:
Hello @rjha-splunk !

Do you already have some information to share?

Thanks!

sbylica-splunk:
We have a (experimental for now) documentation for ECS now, you can check it on develop branch here:
https://splunk.github.io/splunk-connect-for-syslog/develop/gettingstarted/ecs-fargate/

We'd like to gather some feedback on that and push it into official release soon :)

krshx:
hey @sbylica-splunk, Thanks for looking into this and for the update.

I’ve since moved on from that project/company, and as the Fargate option wasn’t available at the time, we ended up using EC2 as the fallback, which was subsequently implemented in production.

If we have a requirement to reimplement SC4S in a future project, I’ll definitely revisit this and update the issue with any findings.

Appreciate you taking the time to look into this.

sbylica-splunk:
[...]

PR #3170, opened 2026-07-29
Feat/ecs fargate support

sbylica-splunk:
Documentation for running SC4S using ECS/Fargate.
Idea first raised in this Github thread: https://github.com/splunk/splunk-connect-for-syslog/issues/2424


sbylica-splunk:
We could also add a pre-check for diskbuffer, kind of like the one we have for HEC connectivity. Not sure if added complexity here is worth it really since we already get this error:
`[2024-04-23T13:46:43.079546] Failed to grab disk-buffer dirlock; filename='/var/lib/syslog-ng/syslog-ng-disk-buffer.dirlock', error='Bad file descriptor (9)'`

docs/gettingstarted/ecs-fargate.md
```diff
@@ -0,0 +1,263 @@
...
+!!! danger "Data-loss without disk buffer"
+    With the disk buffer disabled, SC4S only holds an in-memory queue for the Splunk destination. If Splunk/HEC is unreachable and that queue fills, or the task is stopped or replaced - **buffered events are lost**. This is acceptable for a  proof of concept or for loss-tolerant data, but for production use prefer **Option A (EFS)**.
+
+# Networking and load balancing
+
+In `awsvpc` network mode (required by Fargate) each task gets its own elastic network interface and IP.
+Open the syslog ports on the task's security group: `514/udp`, `514/tcp`, `601/tcp`, `6514/tcp`, and the
+health-check port `8080/tcp`.
+
+(sbylica note: ok we need to talk about it some more during the review!)
```
Kawron: Remove it before merging

docs/resources/ecs/sc4s-task-definition.json
```diff
@@ -0,0 +1,66 @@
...
+      ],
+      "mountPoints": [
+        {
+          "sourceVolume": "sc4s-var",
+          "containerPath": "/var/lib/syslog-ng"
+        }
+      ],
+      "environment": [
+        {"name": "SC4S_DEST_SPLUNK_HEC_DEFAULT_URL",        "value": "https://<SPLUNK_HEC_HOST>:8088"},
+        {"name": "SC4S_DEST_SPLUNK_HEC_DEFAULT_TOKEN",      "value": "<SPLUNK_HEC_TOKEN>"},
```
Kawron, line 46: Could we use secrets here?

docs/gettingstarted/ecs-fargate.md
```diff
@@ -0,0 +1,263 @@
...
+[Load balancing](../architecture/lb/index.md): load balancing syslog has important
+caveats and is not officialy supported by Splunk.
+
+# Run SC4S as a service
+
+Run SC4S as an ECS **service** so ECS keeps the desired number of tasks running and replaces any that
+fail:
+
+```bash
+aws ecs create-service \
```
Kawron, line 195: We have this warning:
```
warning "One task per access point"
    Every task started from a task definition mounts the same access point, so running a service
    with `--desired-count` greater than `1` makes several syslog-ng instances share one buffer
    directory, and collide trying to access the same resources.
```
But we don't specify: `maximumPercent` which according to https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service_definition_parameters.html takes on default value 200% which means that eks will spin another replica before closing pervious one during deployment (if such exists). I think it is better to set `maximumPercent=100` and `minimumHealthyPercent=0`. This will cause downtime though

Kawron (changes requested):

docs/resources/ecs/sc4s-task-definition.json
sbylica-splunk: Technically yes, but I think it's overly complicated for a basic documentation like this. We use the same style in other runbooks like here:
https://splunk.github.io/splunk-connect-for-syslog/main/gettingstarted/docker-compose/
Or here:
https://splunk.github.io/splunk-connect-for-syslog/main/gettingstarted/podman-systemd-general/

docs/gettingstarted/ecs-fargate.md
sbylica-splunk: Oh interesting, will check it out

sbylica-splunk: Ok it makes sense, I've added this to the docs

Kawron (approved):

docs/resources/ecs/sc4s-task-definition.json
Kawron: Okay, we can leave it like this right now. But I think in the future we could improve it. This would be for sure flagged by security team. Also, we use secrets in SC4SNMP as far as I know, so we can adopt this as a standard for all connectors

PR #3172, opened 2026-07-30
Release v3.46.0

sbylica-splunk:
Release pr for v.3.46.0


ansible/Dockerfile
```diff
@@ -1,16 +1,22 @@
...
+    ansible-lint>=6.0.0 --no-cache-dir
 
 WORKDIR /opt
 COPY ./ansible .
 COPY ./charts .
 
 RUN apt update -y \
-    && apt install --no-install-recommends -y build-essential parallel \
-    && apt install --no-install-recommends -y sshpass \
+    && apt install --require-hashes --only-binary :all: \
```
Kawron: Does apt install have `--require-hashes` option? I though this is pip-only argument.

I don't think we have tests for ansible. Was ansible tested when the og PR was merged? I didn't notice it earlier

Kawron (changes requested):

ansible/Dockerfile
sbylica-splunk: Yeah I think you are right, I probably pasted it here by accident.

And no we don't really have tests for this one.

Kawron (approved):

PR #3178, opened 2026-08-03
feat: Document SC4S deployment on GCP

digrajpu:


docs/gettingstarted/gke.md
```diff
@@ -0,0 +1,220 @@
...
+
+If a pod does not start, debug it with:
+
+```bash
+kubectl describe pod {your_pod_name} -n sc4s
+```
+
+# Configure HPA (Horizontal Pod Autoscaler)
+
+The SC4S Helm chart's built-in HPA template uses the deprecated `autoscaling/v2beta1` API which was removed in Kubernetes 1.26. Since all GKE clusters run Kubernetes 1.26 or later, you must create the HPA manually using `autoscaling/v2`.
```
sbylica-splunk: You know this is a good find, but I think that instead of giving a workaround for the user here we should probably just fix the hpa.yaml configuration itself?
Hpa is disabled by default in every deployment but for the user who turns it on it is currently broken

docs/gettingstarted/gke.md
```diff
@@ -0,0 +1,220 @@
...
+
+```bash
+gcloud compute firewall-rules create allow-sc4s-tcp \
+  --network=default \
+  --allow=tcp:514,tcp:601,tcp:6514 \
+  --source-ranges=<your-syslog-sender-ip-ranges>
+
+gcloud compute firewall-rules create allow-sc4s-udp \
+  --network=default \
+  --allow=udp:514,udp:601 \
```
sbylica-splunk: I don't think we have UDP on 601 (by default). Not sure if we should be including this here

docs/gettingstarted/gke.md
```diff
@@ -0,0 +1,220 @@
...
+helm repo update
+```
+
+3. Create a `values.yaml` file with the following GCP-specific configuration:
+
+```yaml
+replicaCount: 2
+
+splunk:
+  hec_url: "https://<your-splunk-host>:8088/services/collector/event"
```
sbylica-splunk: We usually put an example like this: `https://<your-splunk-host>:8088`, it still works the same but probably better for consistency with the rest of our docs

docs/gettingstarted/gke.md
digrajpu: done

docs/gettingstarted/gke.md
digrajpu: done

docs/gettingstarted/gke.md
digrajpu: done

sbylica-splunk (approved):

PR #3179, opened 2026-08-04
Fix/sonarqube issues

sbylica-splunk:
Some additional small fixes to sonarqube issues


package/sbin/entrypoint.sh
```diff
@@ -54,6 +54,7 @@ term_handler() {
   fi
 # 128 + 15 -- SIGTERM on non-existent process (will cause service failure)
   exit 143
+  return 0
```
Kawron, line 57: I don't understand what is the reason for returning 0 here, won't exit kill the process?

Kawron (approved):

package/sbin/entrypoint.sh
sbylica-splunk: It's only there because sonarqube complains about it. It doesn't do anything either way really

PR #3181, opened 2026-08-06
Feat/mcp configurator validator

Kawron:
1. Modify the configuration-tool.sh to work in non-interactive mode
2. Add tool that will be able to run configuration-tool.sh with the correct parameters
3. Add SKILL for using this tool


sc4s_mcp/tools/configuration_tools.py
```diff
@@ -1,10 +1,175 @@
...
+    sc4s_hec_url: str = Field(alias="SC4S_HEC_URL", description="Splunk HEC URL.")
+    sc4s_hec_token: str = Field(
+        alias="SC4S_HEC_TOKEN", description="UUID-form Splunk HEC token."
+    )
+    sc4s_tls_verify: bool = Field(default=True, alias="SC4S_TLS_VERIFY")
+    protocol: Literal["udp", "tcp", "both"] = Field(default="both")
+    mode: Literal["1", "2"] = Field(
+        default="1", description="1 for custom tuning or 2 for hardware-based tuning."
+    )
+    hardware: Literal["16vCPUs", "8vCPUs", "4vCPUs"] = "8vCPUs"
```
digrajpu: Why we have not added field here? every other field uses Field(alias=..., description=...) pattern

configuration-tool.sh
```diff
@@ -283,6 +331,238 @@ apply_hardware_config() {
...
+        return 1
+    fi
+
+    case "${SC4S_MODE:-1}" in
+        1)
+            mode_choice=1
+            ADJUST_FETCH_LIMIT="${SC4S_ADJUST_FETCH_LIMIT:-no}"
+            SC4S_SOURCE_UDP_FETCH_LIMIT="${NI_UDP_FETCH_LIMIT:-1000}"
+            ADJUST_LISTEN_SOCKETS="${SC4S_ADJUST_LISTEN_SOCKETS:-no}"
+            SC4S_SOURCE_LISTEN_UDP_SOCKETS="${NI_UDP_LISTEN_SOCKETS:-2}"
```
sbylica-splunk: I think default here should be 4, no?

configuration-tool.sh
```diff
@@ -283,6 +331,238 @@ apply_hardware_config() {
...
+            CONFIG="$CONFIG
+SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_MEMBUFSIZE=$SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_MEMBUFSIZE"
+        fi
+
+        CONFIG="$CONFIG
+SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_DISKBUFSIZE=$SC4S_DEST_SPLUNK_HEC_DEFAULT_DISKBUFF_DISKBUFSIZE"
+    fi
+}
+
+require_yes_no() {
```
sbylica-splunk, line 434: Does this do anything? In the calls below the return value is ignored. Also we should probably explicitly return 0.

Kawron: At the beginning of the script we use set -e: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

Which means it will return exit when we get 1 value

digrajpu (approved):

sbylica-splunk (approved):

PR #3186, opened 2026-08-11
fix: Normalize env variable for Name and VPS caches

digrajpu:


Kawron (changes requested):
Let's modify the tests to use values other than yes as well. Besides that lgtm

Kawron (approved):

Kawron:
Can yoou rerun the tests, those are failing but it seems not connected to this pr

PR #3198, opened 2026-08-20
fix: Fix NAME cache IPv4/IPv6 key collisions and IPv6 SQLite overflow

digrajpu:
The NAME cache was converting SOURCEIP to a plain integer before storing it as a SQLite key. This caused two bugs:

- Key collision: 0.0.0.1 and ::1 both produce integer 1, so they shared the same cache row and could overwrite each other's hostname.
- IPv6 failure: Most IPv6 addresses require 128 bits, exceeding SQLite's signed 64-bit integer limit, causing an OverflowError on write.

Fix: Replace the integer key with a family-tagged string (v4:192.0.2.2, v6:2001:db8::1) generated by ip_cache_key() using Python's ipaddress module. This guarantees no cross-family collisions and stores safely as a string in SQLite. Existing caches will cold-repopulate under the new key namespace; operators can use SC4S_CLEAR_NAME_CACHE=yes for a one-time cleanup on upgrade.

### DIFF
--- a/package/shared/pylib/parser_source_cache.py
+++ b/package/shared/pylib/parser_source_cache.py
@@ -1,3 +1,4 @@
+import ipaddress
 import traceback
 import socket
 import struct
@@ -45,6 +46,12 @@ def int2ip(addr):
         return int_to_ip6(addr)
 
 
+def ip_cache_key(addr):
+    ip = ipaddress.ip_address(addr)
+    prefix = "v4" if ip.version == 4 else "v6"
+    return f"{prefix}:{ip}"
+
+
 hostdict = str("/var/lib/syslog-ng/hostip")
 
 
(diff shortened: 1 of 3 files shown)


Kawron:
@digrajpu please add description to this PR

Kawron:
Code lgtm, but have you tested it manually? I think it will be worth doing to make sure we don't break anything

digrajpu:
> Code lgtm, but have you tested it manually? I think it will be worth doing to make sure we don't break anything

Did manual testing against a local SC4S container built from this branch with SC4S_USE_NAME_CACHE=yes and SC4S_USE_VPS_CACHE=yes.

- Sent a vpxa learning event - Splunk confirmed sc4s_vendor=vmware, sc4s_product=vsphere
- Inspected SQLite directly - both NAME and VPS cache entries written correctly
- Sent a CEF message without hostname - SC4S resolved source IP to cached hostname via NAME cache ✅
- Sent a generic sshd message - SC4S applied cached vmware/vsphere classification via VPS cache ✅
- Destination stats: d_psc dropped=0, d_vpsc dropped=0 - no errors or drops

Kawron (approved):

PR #3200, opened 2026-08-25
Feat/mcp add skills

Kawron:
Add five SC4S skills for parser creation, configuration, troubleshooting, parser management, and Splunk metadata management. Package the skills as sc4s-plugin for Codex, Claude Code, and Cursor.

PR #3205, opened 2026-08-31
Feat/add sending tool to mcp

Kawron:
* Add `send_syslog_text` tool, that allows to send syslog events to SC4S directly from LLM
* Add skill for using this tool
* Refactor mcp to use `syslog-ng-ctl reload` command, that uses SIGHUP signal, instead of kill signal for operations that don't require changing the env

Please first review:
https://github.com/splunk/splunk-connect-for-syslog/pull/3200/changes


package/sbin/utils.py
```diff
@@ -91,14 +94,34 @@ def restart_syslog_ng():
...
+        remaining = deadline - time.monotonic()
+        if replacement_is_healthy and remaining > 0:
+            return
+        if remaining <= 0:
+            break
+        time.sleep(min(RELOAD_POLL_INTERVAL_SECONDS, remaining))
+
+    raise RuntimeError(
+        f"syslog-ng reload timed out after {RELOAD_TIMEOUT_SECONDS} seconds"
+    )
```
sbylica-splunk, line 124: Why do we need this instead of the previous method?

Kawron: This is just better practice. Syslog-ng reload sends SIGHUP wich in daemons is used to reload conifg. This measn that we won't have to stop the whole process

sc4s_mcp/utils/syslog_sender.py
```diff
@@ -0,0 +1,85 @@
...
+    """Send each non-empty line in *text* to the configured SC4S host."""
+    protocol = protocol.lower()
+    _validate(port, timeout, protocol, framing)
+    payloads = _payloads(text)
+    if framing is None:
+        framing = "raw" if protocol == "udp" else "newline"
+    destination = (_host(), port)
+    started = time.monotonic()
+    sock = None
+    sent = failed = bytes_sent = 0
```
sbylica-splunk, line 52: This "failed" var is never used

sc4s_mcp/utils/syslog_sender.py
```diff
@@ -0,0 +1,85 @@
...
+            sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            sock.settimeout(timeout)
+            for payload in payloads:
+                bytes_sent += sock.sendto(payload, destination)
+                sent += 1
+    finally:
+        if sock:
+            sock.close()
+    return {
+        "status": "ok" if failed == 0 else "partial",
```
sbylica-splunk, line 83: So this else is never true

docs/sc4s-plugin.md
```diff
@@ -0,0 +1,72 @@
...
+
+The repository root is the plugin root, and `skills/` is the single source of skill content. Each client has a small native metadata directory: `.claude-plugin/` for Claude Code, `.codex-plugin/` for Codex, and `.cursor-plugin/` for Cursor. Codex's repository marketplace catalog lives in `.agents/plugins/marketplace.json`.
+
+## Included skills
+
+- `create-parser` creates SC4S syslog-ng parsers from raw log samples.
+- `sc4s-guided-configuration` guides SC4S `env_file` configuration and tuning.
+- `troubleshoot-sc4s` diagnoses health, ingestion, routing, parser, and configuration-job problems.
+- `manage-splunk-metadata` safely manages full-replacement Splunk metadata overrides.
+- `manage-sc4s-parsers` lists, inspects, deploys, updates, and deletes custom parsers with job polling and health verification.
```
sbylica-splunk, line 14: We should probably add send-syslog-text here?

sc4s_mcp/utils/syslog_sender.py
Kawron: fixed

sc4s_mcp/utils/syslog_sender.py
Kawron: fixed

sc4s_mcp/utils/syslog_sender.py
```diff
@@ -0,0 +1,85 @@
+"""Small raw-text syslog sender used by the MCP event tool."""
+
+import os
+import socket
+import time
+from urllib.parse import urlparse
+
+DEFAULT_ALLOWED_PORTS = {514, 601}
```
sbylica-splunk: This is  not used anywhere

docs/sc4s-plugin.md
Kawron: Fixed, though the name of the skill is `sc4s-send-syslog`

sc4s_mcp/utils/syslog_sender.py
Kawron: fixed

tests/test_management_api_integration.py
```diff
@@ -0,0 +1,182 @@
...
+    original_metadata = requests.get(
+        f"{api_url}/config/metadata/splunk", timeout=5
+    )
+    original_metadata.raise_for_status()
+    original_compliance = requests.get(
+        f"{api_url}/config/metadata/compliance", timeout=5
+    )
+    original_compliance.raise_for_status()
+
+    parser_content = f'''\
```
sbylica-splunk: I think it would be better to put these parsers into a file, this looks messy

tests/test_management_api_integration.py
```diff
@@ -0,0 +1,182 @@
...
+            f"{api_url}/config/parser/{parser_stem}", timeout=5
+        )
+        if parser_response.status_code == 200:
+            _delete(api_url, f"/config/parser/{parser_stem}")
+
+        _delete(api_url, "/config/metadata/splunk")
+        if original_metadata.json()["entries"]:
+            _post_json(api_url, "/config/metadata/splunk", original_metadata.json())
+
+        _delete(api_url, "/config/metadata/compliance")
```
sbylica-splunk, line 214: There should probably be an if before othere _delete functions too

sbylica-splunk (changes requested):

tests/test_management_api_integration.py
Kawron: fixed

tests/test_management_api_integration.py
Kawron: fixed and refactored this test to 3 seperate test cases

docs/sc4s-plugin.md
```diff
@@ -0,0 +1,73 @@
...
+
+```shell
+claude plugin marketplace add splunk/splunk-connect-for-syslog
+claude plugin install sc4s-plugin@sc4s-plugins
+```
+
+Reload an active Claude Code session after installation:
+
+```text
+/reload-plugins
```
digrajpu, line 53: What is the use of `/reload-plugins` command?

skills/parser-creator/parser-guide.md
```diff
@@ -0,0 +1,235 @@
...
+
+Use the SC4S parser type appropriate to the format, such as `syslog`, `cef`, or `netsource`. Preserve `<filename>` exactly, including its `.conf` suffix, after selecting it.
+
+## Step 1 — Identify syslog format
+
+- **RFC3164:** `<PRI>TIMESTAMP HOSTNAME PROGRAM: MESSAGE`
+- **RFC5424:** `<PRI>VERSION TIMESTAMP HOSTNAME APP-NAME PROCID MSGID SDATA MESSAGE`
+- **CEF:** `<PRI>TIMESTAMP HOSTNAME CEF:0|Vendor|Product|Version|SigID|Name|Severity|Extensions`
+
+If the format is none of the above, tell the operator it is unsupported and stop.
```
digrajpu: The guide says only RFC3164, RFC5424, and CEF are supported formats and tells operator to stop for anything else. But in `parse_message.md` it tells  use json-parser when logs are JSON-formatted.

docs/sc4s-plugin.md
Kawron: I took it from here: https://code.claude.com/docs/en/plugins#add-skills-to-your-plugin

skills/parser-creator/parser-guide.md
Kawron: Good catch, I will fix it

sc4s_mcp/utils/syslog_sender.py
```diff
@@ -0,0 +1,92 @@
...
+            for payload in payloads:
+                framed = (
+                    f"{len(payload)} ".encode("ascii") + payload
+                    if framing == "octet-counting"
+                    else payload + b"\n"
+                )
+                try:
+                    sock.sendall(framed)
+                except OSError:
+                    failed += 1
```
digrajpu, line 109: Here we can optimize the code. when the first TCP OSError occur we can break the loop
```
except OSError:
    failed += len(payloads) - sent  # count all remaining as failed
    break
```

Kawron: fixed

Kawron: On the second though I reverted that change. I think the optimization is minuscule, and there is no reason to cancel next packets when the first won't work.

package/sbin/utils.py
```diff
@@ -91,14 +94,34 @@ def restart_syslog_ng():
...
+        timeout=10,
+    )
+    if result.returncode != 0:
+        error = result.stderr.strip() or result.stdout.strip()
+        raise RuntimeError(f"syslog-ng reload failed: {error}")
+    deadline = time.monotonic() + RELOAD_TIMEOUT_SECONDS
+    while time.monotonic() < deadline:
+        current_pids = _get_syslog_ng_pids()
+        replacement_is_healthy = (
+            bool(current_pids == previous_pids) and _syslog_ng_is_healthy()
```
sbylica-splunk: Should this condition be this strict? If I understand correctly if **any** of the pids are different we will return a timeout, even if there are actually healthy instances.

package/sbin/utils.py
```diff
@@ -91,14 +94,34 @@ def restart_syslog_ng():
...
 
-def rollback_env():
-    """Restore env_file from backup and restart syslog-ng."""
-    if BACKUP_FILE.exists():
-        shutil.copy(BACKUP_FILE, ENV_FILE)
-        try:
-            restart_syslog_ng()
-        except Exception:
-            logger.exception("Rollback also failed")
+def reload_syslog_ng():
```
sbylica-splunk, line 97: We kind of use reload as a default now, but what if the process hangs? There should probably be workflow like this at least:
reload -> if it doesn't work, restart instead - like before

sc4s_mcp/Dockerfile.dockerignore
```diff
@@ -13,6 +13,7 @@
 !docs/
 !package/shared/addons/
 !skills/parser-creator/
+!skills/sc4s-configurator/
```
sbylica-splunk: Why do we skip this?

sbylica-splunk: Also are we including the new directories into docker image? I don't see a copy command in the dockerfile for some of the directories, like skills/...

sc4s_mcp/utils/syslog_sender.py
```diff
@@ -0,0 +1,92 @@
...
+    hostname = urlparse(os.getenv("SC4S_API_URL", "")).hostname
+    if not hostname:
+        raise ValueError("SC4S_API_URL must contain a hostname")
+    return hostname
+
+
+def _validate(port, timeout, protocol, framing):
+    if not isinstance(port, int) or isinstance(port, bool) or not 1 <= port <= 65_535:
+        raise ValueError("port must be an integer between 1 and 65535")
+    allowed = os.getenv("SC4S_MCP_ALLOWED_SYSLOG_PORTS", "514,601")
```
sbylica-splunk: We should probably check this line too: if SC4S_MCP_ALLOWED_SYSLOG_PORTS is malformed it may raise an unrelated error in the if condition below.

sc4s_mcp/utils/syslog_sender.py
```diff
@@ -0,0 +1,92 @@
...
+    if not payloads:
+        raise ValueError("No syslog events found in raw text input")
+    if any(len(payload) > MAX_EVENT_BYTES for payload in payloads):
+        raise ValueError("event exceeds maximum size of 65507 bytes")
+    return payloads
+
+
+def send_text(*, text, protocol, port, framing=None, timeout=5.0):
+    """Send each non-empty line in *text* to the configured SC4S host."""
+    protocol = protocol.lower()
```
sbylica-splunk, line 88: This should happen after the validation

sc4s_mcp/utils/syslog_sender.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _host():
+    hostname = urlparse(os.getenv("SC4S_API_URL", "")).hostname
+    if not hostname:
+        raise ValueError("SC4S_API_URL must contain a hostname")
+    return hostname
+
+
+def _validate(port, timeout, protocol, framing):
+    if not isinstance(port, int) or isinstance(port, bool) or not 1 <= port <= 65_535:
```
sbylica-splunk, line 41: isinstance(port, bool) seems excessive here? it should already be covered by not isinstance(port, int) right?

sc4s_mcp/utils/syslog_sender.py
```diff
@@ -0,0 +1,92 @@
...
+            sock = socket.create_connection(destination, timeout=timeout)
+            for payload in payloads:
+                framed = (
+                    f"{len(payload)} ".encode("ascii") + payload
+                    if framing == "octet-counting"
+                    else payload + b"\n"
+                )
+                try:
+                    sock.sendall(framed)
+                except OSError:
```
sbylica-splunk: Can we add some better error handling here?
https://docs.python.org/3/library/socket.html#exceptions

tests/api/test_utils.py
```diff
@@ -127,49 +127,102 @@ def test_restart_fails_immediately_when_kill_fails(mock_run):
...
+            capture_output=True,
+            text=True,
+            timeout=2,
+        ),
+    ]
+    mock_sleep.assert_not_called()
+
+
+@patch("utils.subprocess.run")
+def test_reload_reports_error_from_stdout_when_stderr_is_empty(mock_run):
```
sbylica-splunk: There are few more branches for restart (restart_times_out, restart_rejects etc.), can we test reload path in similar situations too?

package/sbin/utils.py
Kawron: We do have this:
```
 while time.monotonic() < deadline:
        current_pids = _get_syslog_ng_pids()
        replacement_is_healthy = (
            bool(current_pids == previous_pids) and _syslog_ng_is_healthy()
        )
        remaining = deadline - time.monotonic()
        if replacement_is_healthy and remaining > 0:
            return
        if remaining <= 0:
            break
        time.sleep(min(RELOAD_POLL_INTERVAL_SECONDS, remaining))

    raise RuntimeError(
        f"syslog-ng reload timed out after {RELOAD_TIMEOUT_SECONDS} seconds"
    )
```

sc4s_mcp/utils/syslog_sender.py
Kawron: Why? What's wrong with allowing values as `UDP` or `Udp`?

sc4s_mcp/utils/syslog_sender.py
Kawron: Bool is subclass of int in python
```
8:52:13 splunk-connect-for-syslog $ python3         git:(feat/add-sending-tool-to-mcp*)
Python 3.11.9 (v3.11.9:de54cf5be3, Apr  2 2024, 07:12:50) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> a = True
>>> print(isinstance(a, int))
True
>>> 
```

package/sbin/utils.py
Kawron: After consideration it doesn't have to be this strict. I will change that.

Because this sends SIGHUP the syslog-ng should restart itself and the pids should be the same, but I don't think we need to check that

tests/api/test_utils.py
Kawron: I added two more negative examples

sc4s_mcp/Dockerfile.dockerignore
Kawron: We don't skip it, we exclude it from dockerginore file so it's not ignore, see the * wildcard. But actually it is not needed in the mcp server at all so we can delete it

digrajpu (approved):

PR #3219, opened 2026-09-08
fix: Add GCP deployment information to Experimental section

digrajpu:


docs/experiments.md
```diff
@@ -6,6 +6,17 @@ SC4S can now be deployed on AWS ECS/Fargate and Google Kubernetes Engine (GKE).
...
 These guides are in an experimental phase now, and should be treated as such - it's likely they will be changed / improved in the future.
 
+### GCP (Google Cloud Platform)
+SC4S can now be deployed on Google Cloud Platform (GCP) using Google Kubernetes Engine (GKE).
+See the [GKE getting started guide](gettingstarted/gke.md) for setup steps.
+This guide is in an experimental phase and should be treated as such - it's likely it will be changed / improved in the future.
+
+**GCP-specific notes:**
+
+- GCP automatically provisions a Passthrough Network Load Balancer (L4) when `service.type: LoadBalancer` is set — no manual load balancer setup required, unlike EKS.
```
Kawron: I would remove the part: 
```
no manual load balancer setup required, unlike EKS.
```
I don't think we have to compare those two. It is unlikely one will choose their cloud infrastracture based on diffuculty of SC4S installation.

It also reads as LLM generated text

Kawron (changes requested):

docs/experiments.md
digrajpu: Ohk I will update it and yes, I used a LLM to help rephrase the wordings.

Kawron: Ok, there is no issue with using LLM. It just sounds weird and synthetic. LLM assumed that you want to compare eks to gks, but human being lacks this context, so this comment seems out of place. But that's just my though, on the writing style

Kawron (approved):

sbylica-splunk (approved):

PR #3225, opened 2026-09-14
Chore/update troubleshooting docs

Kawron:
Add additonal sections on troubleshooting docs. I've added how to read syslog-ng-ctl stats and how to check for tcp backpressure


docs/troubleshooting/troubleshoot_resources.md
```diff
@@ -26,6 +26,144 @@ podman volume create splunk-sc4s-var
...
+The additional `ss -i` output can include the round-trip time (`rtt`),
+retransmission timeout (`rto`), retransmission backoff, congestion window
+(`cwnd`), acknowledged bytes, and estimated send rate. A
+`timer:(persist,...)` entry is strong evidence that the remote endpoint
+advertised a zero receive window and the local host is sending window probes.
+Increasing retransmissions alone can indicate packet loss rather than
+application backpressure.
+
+If `tshark` is installed, use packet analysis to confirm zero-window events,
+receiver-window-full conditions, or retransmissions:
```
digrajpu: Here are we missing tshark command?

Kawron: I didn't want to give tshark commands, because those are a little bit more complicated, and that's another thing to that we would have to maintain. E.g. Customer could come to us and say: "the command you provided is not working, can you help us with it".

So maybe safer option is to just remove this paragraph

digrajpu (approved):

